PR #4, opened 2019-11-21
Toolbar

Joozty:


components/header/index.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import './index.scss'
+import Logo from '../logo'
+import Header from './Header'
+import HeaderItem from './HeaderItem'
+import Avatar from '../avatar'
+import SearchBar from '../searchbar'
```
viktor-yakubiv: This is actually why I asked you to send an early work. You have too many dependencies here. Header should not contain everything. There is a business logic and the view rendering in the same time. Try to do something like following:

```jsx
const AppBar = ({ children, className, tag: Tag = 'div', ...restProps}) => (
  <Tag 
    className={combine(modulize('app-bar', classNames), className)}
    {...restProps}
  >
    {children}
  </Tag>
)

const Brand = ({ ... }) => ...
AppBar.Brand = Brand

export ...

...
...


const DashboardAppBar = ({ user }) => (
  <AppBar className="dashboard-app-bar">
    <AppBar.Brand>Dashboard</AppBar.Brand>
    ...
    <AppBar.Avatar src={user.avatarUrl} alt={`${user.name}'s avatar`} />
  </AppBar>
)
```

The first part contains pure components that are as simple as possible and may freely go to the library. The last one is dashboard-dependent and stays in the dashboard.

viktor-yakubiv:
The Dashboard has 2-level navigation. The first one (level) appears at the application bar. Second level appears at the sidebar.

In both cases a _navigation_ means having titled (containing a heading) `<nav>` element with a list inside.

To simplify primary navigation we add a search that works on the client side. However, behind that it's a custom _select_ element based on `<ul>`.

components/header/index.jsx
Joozty: Well, my initial idea was to move `Header`, `HeaderItem`, `Avatar`, `SearchBar` to the library once it's done and in `components/header/` would stay only `index.js`. I guess it's mostly the same what you are proposing.

> ```js
> const Brand = ({ ... }) => ...
> AppBar.Brand = Brand
> ```

I do like to have it like that. I wanted to have it like `import {Header, Brand, HeaderItem} from '@oacore/design'`. Your suggestion seems much better. Thanks.


What I do not understand are functions `combine` & `modulize`. Where did you get them from?

viktor-yakubiv: `combine` & `modulize` are abstract for this moment. I wanted to add an implementation as utilities into the library.

The first one may be [@JedWatson/classNames](https://github.com/JedWatson/classnames) but I do not really like that its name does not say what it does.

The second one is a simple map over an object.

components/layout/index.scss
```diff
@@ -0,0 +1,81 @@
...
+    > * {
+      padding-left: 1rem;
+      padding-right: 1rem;
+    }
+  }
+
+  .layout-sidebar {
+    position: absolute;
+    background: gray;
+    left: -$sidebarWidth;
```
viktor-yakubiv: It's better to use `transform` because it has better performance and implicitly tells browser that the element should be painted in a separate layer because it might change.

components/logo/logo.scss
```diff
@@ -0,0 +1,22 @@
...
+  color: #2a3840;
+  padding-top: 0.40625rem;
+  padding-bottom: 0.40625rem;
+  font-size: 1.125rem;
+
+  &:hover {
+    color: var(--primary);
+  }
+
+  & > img {
```
viktor-yakubiv: You can nest it implicitly without `&` here.

It's better to avoid direct child nesting (`>`) where it's possible. It might happen you want to add a wrapper around `img` so then you will have a line more changes.

components/searchbar/index.jsx
```diff
@@ -0,0 +1,12 @@
+import './index.scss'
+import React from 'react'
+
+const SearchBar = () => {
```
viktor-yakubiv: We have different imagination about this part. It will cause further discussion but it's fine for now 🙂

components/sidebar/index.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import './index.scss'
+
+const Sidebar = React.memo(({ className }) => {
+  return (
+    <nav className={className}>
+      <ul>
+        <li>
+          <h3>Overview</h3>
```
viktor-yakubiv: It's not semantically correct to use headings here. Document outline might look weird because of that.

design/hamburger/index.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import './index.scss'
+
+const Hamburger = React.memo(({ onClick }) => (
```
viktor-yakubiv: This is interesting. If we do it configurable it would be amazing component 🙂

design/header/index.jsx
```diff
@@ -0,0 +1,15 @@
...
+
+const Header = React.memo(({ className, children }) => (
+  <header className={classNames('header', className)}>{children}</header>
+))
+
+const HeaderItem = React.memo(({ className, children }) => (
+  <div className={classNames('header-item', className)}>{children}</div>
+))
+
+Header.HeaderItem = HeaderItem
```
viktor-yakubiv: Would not be better to have `Header.Item = HeaderItem`?

package.json
```diff
@@ -40,6 +41,7 @@
     "dev": "next dev",
     "build": "next build",
     "start": "next start",
-    "lint": "eslint --ignore-path .gitignore --ext .js,.jsx ."
+    "lint": "eslint --ignore-path .gitignore --ext .js,.jsx .",
+    "format": "eslint --ignore-path .gitignore --fix --ext .js,.jsx ."
```
viktor-yakubiv: You put the `format` back. I tried to avoid it because with yarn it's possible to use it as simple as `yarn lint --fix` but `npm` does things completely weird. Although, `npm run link -- --fix` could work (I did not test it enough times).

pages/_app.jsx
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  render() {
+    const { Component, pageProps } = this.props
+
+    return (
+      <Layout>
+        {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+        <Component {...pageProps} />
+      </Layout>
```
viktor-yakubiv: I expected to bootstrap all application behaviour and layout here. Any ideas/objectives?

pages/_app.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+    return { pageProps }
+  }
+
+  render() {
+    const { Component, pageProps } = this.props
+
+    return (
+      <Layout>
+        {/* eslint-disable-next-line react/jsx-props-no-spreading */}
```
viktor-yakubiv: I prefer spread rest props. It's just convenient.

When you write a wrapper around an element, let's say `<input>` you often do not even know it's attributes. Also, there is a wide range of global attributes such as id, lang, class, aria-* that a user might want to pass but component does not allow it. `restProps` is a way to provide usability and be safe it development. You extract what you want but pass everything else down to the child.

pages/index.jsx
```diff
@@ -1,31 +1,8 @@
 import React from 'react'
-import Head from 'next/head'
+import './index.scss'
```
viktor-yakubiv: CSS/SCSS — it's a matter of discussion but we may mix if you want to.

design/header/index.scss
```diff
@@ -0,0 +1,16 @@
+.header {
+  background-color: #f8f9fa !important;
```
viktor-yakubiv: This important does not look really good.

components/layout/index.scss
```diff
@@ -0,0 +1,81 @@
+$sidebarWidth: 10rem;
+
+.layout {
+  grid-template-rows: min-content auto;
+  position: relative;
+  min-height: 100vh;
```
viktor-yakubiv: I would prefer percents for safety. For example, in Reader's case if you embed the Reader (with this property) into a modal window you will experience issues.

components/header/index.scss
```diff
@@ -0,0 +1,11 @@
+.header-item-avatar-wrapper {
+  justify-content: flex-end;
+}
+
+.header-item-search-bar-wrapper {
+  padding: 0;
+  justify-content: center;
+  flex-basis: 60%;
```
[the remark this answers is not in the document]
Joozty: Probably leftover

package.json
Joozty: I wondered where did it go. Sorry I get it now.

components/header/index.scss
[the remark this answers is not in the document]
Joozty: No leftover. It's because the search bar should be significant bigger than the avatar and logo.

pages/_app.jsx
Joozty: I followed the same pattern as is in @oacore/next repository. I kind of like it as is but don't mind to move layout directly here.

components/header/index.jsx
```diff
@@ -0,0 +1,22 @@
...
+import './index.css'
+import Logo from '../logo'
+import Header from '../../design/header'
+import Avatar from '../../design/avatar'
+import SearchBar from '../searchbar'
+
+export default React.memo(({ className }) => {
+  return (
+    <Header className={className}>
+      <Header.Item className="header-area-brand">
```
viktor-yakubiv: Why do you need `Header.Item`?

components/head/index.jsx
```diff
@@ -0,0 +1,37 @@
...
+      rel="icon"
+      sizes="any"
+      type="image/svg+xml"
+      href="/favicon/favicon.svg"
+    />
+    <style>
+      {`
+          html,
+          body,
+          #__next {
```
viktor-yakubiv: I wonder if we can pass another class name to `NextContainer`

components/header/index.css
```diff
@@ -0,0 +1,16 @@
+.header-area-user {
+  justify-content: flex-end;
+}
+
+.header-item.header-area-navigation {
```
viktor-yakubiv: Looks weird in comparison to selector above. Write either with `.header-item` or without it but be consistent. Unless you cannot, write a comment please.

I would prefer without because it's clearer unless you want to increase specifity to override declarations without `!important`.

Just want to clarify 🧐

components/header/index.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import './index.css'
+import Logo from '../logo'
+import GlobalHeader from '../../design/header'
+import Avatar from '../../design/avatar'
+import SearchBar from '../searchbar'
+
+const Header = React.memo(({ className }) => {
+  return (
+    <GlobalHeader className={className}>
```
viktor-yakubiv: I would call it `AppBar` just because it's more meaningful and common for Material Design. Also, the opposite works — AppBar should not always be a header and vice versa.

components/layout/index.jsx
```diff
@@ -0,0 +1,20 @@
...
+import Head from '../head'
+import './index.css'
+import Sidebar from '../sidebar'
+
+const Layout = ({ children }) => {
+  return (
+    <>
+      <Head />
+      <div className="layout">
+        <Header className="layout-bar fixed" />
```
viktor-yakubiv: You removed mobile behaviour at all, right?

components/searchbar/index.jsx
```diff
@@ -0,0 +1,12 @@
+import './index.css'
+import React from 'react'
```
viktor-yakubiv: No, it should not be like this.

Global modules go first, local ones second and stylesheets go the last because it's not even JS.

components/layout/index.jsx
Joozty: yes will add it in next PR

components/header/index.jsx
Joozty: Just to be consistent. I wanted to have it like only HederItem can be direct child of the Header component.

components/header/index.css
Joozty: I wanted to avoid !important

components/header/index.scss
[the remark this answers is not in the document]
viktor-yakubiv: Just seen the answer 😁

If so, sidebar should grow two times faster then the rest, i.e. `flex: 2 0 2px` would be enough.

Logo should not grow at all, so `flex: 0` for it.

PR #13, opened 2019-11-27
Mobile header

Joozty:
@viktor-yakubiv Would you mind if I play with mobile header animation later on? I want to deliver MWE as fast and possible. We can tweak details then. Unless you know how to do the animation similar easy without involving much javascript.


components/layout/index.jsx
```diff
@@ -1,20 +1,36 @@
...
 const Layout = ({ children }) => {
+  const [isSidebarVisible, toggleSidebarVisibility] = useState(false)
+
+  const handleCLick = e => {
+    const { name, value } = e.target
+    if (['sidebar.show', 'sidebar.close'].includes(`${name}.${value}`))
+      toggleSidebarVisibility(!isSidebarVisible)
+  }
+
+  // TODO: Ask Viktor how aria should be handled
```
Joozty: @viktor-yakubiv Could you please help me out here? DIV element bellow has onClick listener and I don't know which aria role should be used. 😕

components/header/index.css
```diff
@@ -8,9 +8,66 @@
...
 .header-item.header-area-brand {
   flex-grow: 0;
   flex-basis: 11.5rem;
 }
+
+.menu-toggle {
+  display: none;
+  height: 100%;
+  justify-content: center;
+  flex-direction: column;
```
viktor-yakubiv: Why flex?

components/header/index.css
```diff
@@ -8,9 +8,66 @@
...
+  border: none;
+  background: none;
+  cursor: pointer;
+}
+
+.menu-toggle > span {
```
viktor-yakubiv: I would still prefer `.bar` even if you use non-semantic HTML element

components/layout/index.jsx
viktor-yakubiv: Toggle button should have `aria-haspopup="true" aria-controls="sidebar-id" aria-expanded="false"`. The last one should change when sidebar toggles.

The most tricky one is ID in the current application structure. That is why I would do layout in a single file somehow. Otherwise you have to pass this ID to header and then pass it to the bottom.

components/layout/index.jsx
```diff
@@ -1,20 +1,36 @@
...
+  /* eslint-disable */
   return (
     <>
       <Head />
-      <div className="layout">
+      <div className="layout" onClick={handleCLick}>
         <Header className="layout-bar fixed" />
```
viktor-yakubiv: I would rather add listener to `Header` to make it as close as possible to the toggle button but not passing it directly.

React does event delegation on mounting point so here it's just matter of taste.

Also, you have a typo in `CLick`  and would rather prefer having more concrete name, i.e. `handleSidebarToggle`, unless you do really complex stuff in the event listener.

components/layout/index.jsx
```diff
@@ -1,20 +1,36 @@
...
 import Header from '../header'
 import Head from '../head'
 import './index.css'
 import Sidebar from '../sidebar'
 
 const Layout = ({ children }) => {
+  const [isSidebarVisible, toggleSidebarVisibility] = useState(false)
+
+  const handleCLick = e => {
+    const { name, value } = e.target
```
viktor-yakubiv: `e` may be confusing, I might mean either `event` or `error`, and both are quite common everywhere. I would prefer call `event` actually `event`.

components/sidebar/index.jsx
```diff
@@ -1,9 +1,20 @@
 import React from 'react'
 import './index.css'
+import Logo from '../../design/logo'
 
 const Sidebar = React.memo(({ className }) => {
   return (
     <nav className={className}>
+      <div className="branding">
```
viktor-yakubiv: It's actually a `<header>` and maybe even an AppBar but I am not sure yet 🙂

We have something similar in CORE Discovery browser-extension. Maybe you have better ideas.

CORE Discovery Sidebar

pages/index.css
```diff
@@ -26,6 +26,8 @@
   --breakpoint-lg: 992px;
   --font-family-sans-serif: Roboto, -apple-system, BlinkMacSystemFont, "Segoe UI", "Helvetica Neue", Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
   --font-family-monospace: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
+  --mobile-view-breakpoint: 800px;
+  --sidebar-width: 13rem;
```
viktor-yakubiv: It's actually `12rem`. One extra is for `padding` for the rest of the page.

components/sidebar/index.css
```diff
@@ -0,0 +1,28 @@
...
+  justify-self: flex-end;
+  position: relative;
+  background: none;
+  border: none;
+  cursor: pointer;
+}
+.close:hover {
+  opacity: 0.3
+}
+.close:before, .close:after {
```
viktor-yakubiv: You could use an icon but it's up to you.

components/layout/index.css
```diff
@@ -33,7 +34,46 @@
...
   padding-top: 4.5rem;
 }
+
+.layout-sidebar > .branding {
+  display: none;
+  background: #FAFAFA;
+  padding: 1rem;
+}
+
+@media only screen and (max-width: 800px) {
```
viktor-yakubiv: You have a variable below but perhaps it does not work. There is `@cutom-media` rule for such things.

components/header/index.css
```diff
@@ -8,9 +8,66 @@
   flex-basis: 60%;
   align-self: stretch;
   height: inherit;
+  width: 100%;
+  transition: transform 0.3s ease-in, border 0s ease-in 0.3s;
+  will-change: transform;
```
viktor-yakubiv: I do not really get where you do change transform.

components/layout/index.css
```diff
@@ -33,7 +34,46 @@
...
+  .layout-sidebar {
+    left: calc(var(--sidebar-width) * -1);
```
viktor-yakubiv: I would rather have `transform` here.

```css
left: 0;
transform: translateX(-100%);
```

and on toggle:
```css
transform: translateX(0); /* can be simplified to 0 */
```

components/layout/index.css
Joozty: Variables in @media queries don't work as you said.

I need to set up postcss properly. It's not gonna happen in this PR. I am still waiting once the design package set up is done then I will inspire (copy & paste 😄 ) from there.

components/layout/index.jsx
Joozty: ok fair enough

components/layout/index.css
viktor-yakubiv: Good point ☝️

components/header/index.css
Joozty: Well, I don't think about it much. I just use the first thing that works for me. If you have better css styling please share code snippet. 🙂

components/sidebar/index.css
Joozty: This approach doesn't require image download. None that it matters. Maybe I will switch to icon later on.

components/header/index.css
Joozty: It's here. https://github.com/oacore/dashboard/blob/mobile-view/components/header/index.css#L56 However the animation doesn't work properly. I will need more time to make it look better.

components/header/index.css
viktor-yakubiv: It's good enough for now

components/header/index.css
viktor-yakubiv: It's more about structure. I will create better design for you

components/sidebar/index.jsx
Joozty: I am not sure if I understand it. Like this element should be in nav instead it should be in `header`/`appBar` component?

viktor-yakubiv: No, not really. You have no restriction to count of `<header>`, `<nav>`, `<section>` and other semantic elements on the page.

Also, you have no restriction where to use `AppBar` component. You can do it everywhere it fits your needs.

On the other side, I feel like this one might be a fake DRY so I would not change it for this time. Sorry, I am confusing.

components/layout/index.jsx
Joozty: Thanks, unfortunately, there were other a11-js warnings. I left there TODO.

PR #14, opened 2019-11-28
Sidebar draft

Joozty:
@viktor-yakubiv Don't pay much attention to design. It's ugly. There haven't been applied many css styles. It's mostly for making subpages working. Also, I copied & pasted the icon component from Reader. I believe we will migrate to design package soon.


components/sidebar/index.jsx
```diff
@@ -15,12 +17,47 @@ const Sidebar = React.memo(({ className }) => {
...
-        <li>Overview</li>
-        <li>Data</li>
-        <li>Statistics</li>
-        <li>Plugins</li>
-        <li>Settings</li>
+      <ul className="sidebar-navigation">
+        <li>
+          <Link href="/">
+            <a href="/">
+              <Icon iconType="overview" />
```
viktor-yakubiv: I am not sure about having an explicit `Icon`. On the other hand we do not have any way to use icons in CSS. I will figure out something.

components/layout/index.css
```diff
@@ -1,4 +1,5 @@
 .layout {
+  margin-right: -1rem;
```
viktor-yakubiv: If it's top, kind of root container, this might cause horizontal scrolling.

components/sidebar/index.css
```diff
@@ -26,3 +26,43 @@
 .close:after {
   transform: rotate(-45deg);
 }
+
+ul.sidebar-navigation,
+ul.sidebar-navigation li > a {
```
viktor-yakubiv: I would prefer use class to style this. It's more convenient. If you then go to `<button>` it would be saver.

pages/_app.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import NextApp from 'next/app'
-
```
viktor-yakubiv: I would keep it to explicitly divide global and local but this is up to you.

Joozty: I agree. Please check this [PR](https://github.com/oacore/configs/pull/11). After we merge it I will do proper changes here. Unfortunately, this rule doesn't apply on assets imports. I propose we should have .css/.svg import as a separate 4th group.

Joozty:
@viktor-yakubiv thanks for the review. 🙂

PR #17, opened 2019-12-05
Primary navigation

viktor-yakubiv:
I started primary navigation here. I should be based on an accessible custom select, so it's quite tricky to implement it right.

I was consulting at [WebAxe](https://www.webaxe.org/accessible-custom-select-dropdowns/), especially on the [first article](https://uxmastery.com/anatomy-of-an-accessible-auto-suggest/) from the list.

### DIFF
--- a/design/select/index.jsx
+++ b/design/select/index.jsx
@@ -172,21 +172,19 @@ const Select = ({
   )
 }
 
-const SelectOption = ({ id, children, value, selected, ...restProps }) => {
-  return (
-    <li
-      id={id}
-      role="option"
-      aria-selected={selected}
-      className={classNames({
-        [styles.optionSelected]: selected,
-      })}
-    >
-      <button type="button" value={value} tabIndex="-1" {...restProps}>
-        {children}
-      </button>
-    </li>
-  )
-}
+const SelectOption = ({ id, children, value, selected, ...restProps }) => (
+  <li
+    id={id}
+    role="option"
+    aria-selected={selected}
+    className={classNames({
+      [styles.optionSelected]: selected,
+    })}
+  >
+    <button type="button" value={value} tabIndex="-1" {...restProps}>
+      {children}
+    </button>
+  </li>
+)
 
 export default Select
--- a/design/select/index.jsx
+++ b/design/select/index.jsx
@@ -13,6 +13,7 @@ const generateId = () =>
 const KEYS = {
   ESC: 27,
   ENTER: 13,
+  LEFT: 37,
   UP: 38,
   RIGHT: 39,
   DOWN: 40,
@@ -46,11 +47,6 @@ const Select = ({
       case KEYS.ESC:
         inputRef.current.blur()
         toggleShowSuggestions(false)
-        break
-
-      case KEYS.RIGHT:
-        inputRef.current.blur()
-        onSelectionChange(activeSuggestion)
         break
 
       case KEYS.ENTER:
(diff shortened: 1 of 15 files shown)


viktor-yakubiv:
This is an initial work. @Joozty it would be nice if you can continue it. We should not implement custom select for this moment and can keep this as simple as possible being a list with links.

viktor-yakubiv:
Let's do only pure select with a pretty label on top of it. Also, let's work only with data-provider name only, without organisation name appended.

Joozty:
@viktor-yakubiv This is initial work. CSS needs to be tweaked more but I would prefer to do it in new PR. (It would be even better if you play with CSS, but not mandatory. 🙂)

design/select/index.jsx
```diff
@@ -0,0 +1,190 @@
...
+    switch (event.which) {
+      case KEYS.ESC:
+        inputRef.current.blur()
+        toggleShowSuggestions(false)
+        break
+
+      case KEYS.RIGHT:
+        inputRef.current.blur()
+        onSelectionChange(activeSuggestion)
+        break
```
viktor-yakubiv: As a user I must be able to navigate my input in both directions. This disables such functionality.

Joozty: What do you mean?

Arrows UP/Down are for navigation. Right arrow and Enter is for selecting option.

design/select/index.jsx
```diff
@@ -0,0 +1,190 @@
...
+        if (suggestions.length === 0) return
+
+        if (activeSuggestion.id === undefined) {
+          setActiveSuggestion(suggestions[0])
+          return
+        }
+
+        pos = pos >= suggestions.length ? suggestions.length : pos + 1
+        setActiveSuggestion(suggestions[pos])
+        break
```
viktor-yakubiv: It can be shorter:

```js
case KEYS.UP:
case KEYS.DOWN:
  event.preventDefault()
  event.stopPropagation()
  if (suggestions.length === 0) return

  if (activeSuggestion.id === undefined) {
    setActiveSuggestion(suggestions[0])
    return
  }

  const direction = event.which - KEYS.DOWN + 1 // either -1 or 1
  pos = (pos + direction) % suggestions.length
  setActiveSuggestion(suggestions[pos])
  break
```

Joozty: Fixed here https://github.com/oacore/dashboard/pull/17/commits/dc372d560a95150c7db3d96eca3f4219b97c64ec Thanks

design/select/index.jsx
viktor-yakubiv: Only Enter should do it. Right arrow should navigate the input.

Joozty:
@viktor-yakubiv Please have a final look and feel free to merge.

design/select/styles.css
```diff
@@ -0,0 +1,56 @@
+:root {
+  --select-background: var(--gray-50);
+}
+
+/* TODO: Better name, it's a div shown on top of input when not hovered */
+.input-container {
```
viktor-yakubiv: `overlay` would work I guess

design/select/index.jsx
```diff
@@ -62,6 +58,8 @@ const Select = ({
         inputRef.current.blur()
         break
 
+      case KEYS.RIGHT:
+      case KEYS.LEFT:
```
viktor-yakubiv: I guess we misunderstood each other. I expected you to remove left and right at all. When you press left or right arrow key, text cursor should go (iterate) over text (user input). It should not be prevented.

PR #18, opened 2019-12-07
@oacore/design & mobX

Joozty:
@viktor-yakubiv Last commit may be particularly interesting for you. Maybe you even will know a solution for it. :)


pages/plugins.css
```diff
@@ -0,0 +1,4 @@
+.plugins {
+  display: flex;
+  flex-direction: column;
```
viktor-yakubiv: Why did not bare blocks satisfy you?

pages/index.css
```diff
@@ -1,48 +1,8 @@
...
-}
-*,
-*::before,
-*::after {
-  box-sizing: inherit;
-}
-
 body,
 html {
   display: block;
```
viktor-yakubiv: This is part of design too.

pages/index.css
```diff
@@ -1,48 +1,8 @@
...
-  --breakpoint-xs: 0;
-  --breakpoint-sm: 576px;
-  --breakpoint-md: 768px;
-  --breakpoint-lg: 992px;
-  --font-family-sans-serif: roboto, -apple-system, blinkmacsystemfont,
-    'Segoe UI', 'Helvetica Neue', arial, sans-serif, 'Apple Color Emoji',
-    'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
-  --font-family-monospace: sfmono-regular, menlo, monaco, consolas,
-    'Liberation Mono', 'Courier New', monospace;
   --mobile-view-breakpoint: 800px;
```
viktor-yakubiv: Consider using [@custom-media](https://github.com/postcss/postcss-custom-media)

pages/_app.jsx
```diff
@@ -2,25 +2,35 @@ import React from 'react'
...
 import Layout from '../components/layout'
+import { initializeData, GlobalProvider } from '../store'
+
+import '@oacore/design/lib/index.css'
 import './index.css'
 
 class App extends NextApp {
   static async getInitialProps({ Component, ctx }) {
     let pageProps = {}
+    const initialStoreData = initializeData()
```
viktor-yakubiv: I feel this should be `await`

store/index.jsx
```diff
@@ -0,0 +1,47 @@
...
+    ...initialData,
+  }
+}
+
+export const withGlobalStore = Component => {
+  return props => (
+    <GlobalContext.Consumer>
+      {value => {
+        const ObservableComponent = observer(Component)
+        return <ObservableComponent {...props} store={value} />
```
viktor-yakubiv: I would change the order. If somebody redefines `store` prop, it's by purpose, IMHO.

pages/plugins.jsx
```diff
@@ -1,15 +1,53 @@
 import React from 'react'
 
+import Plugin from '../components/plugin'
+import pluginsClassNames from './plugins.css'
+
+const integrations = [
+  {
+    name: 'Analytics',
+    abstract:
```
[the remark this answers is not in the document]
Joozty: Nothing better came into my mind. Maybe `shortDescription` would we better?

pages/plugins.jsx
```diff
@@ -1,15 +1,53 @@
...
+      <p>Integrate our powerful services into your website.</p>
+      <div className={pluginsClassNames.plugins}>
+        {integrations.map(i => (
+          <Plugin
+            key={i.name}
+            name={i.name}
+            abstract={i.abstract}
+            description={i.description}
+          />
+        ))}
```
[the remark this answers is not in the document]
Joozty: I am not a big fan of spread syntax because of potential bugs.

I still haven't decided whether we should have separate components for every plugin or rather to have one general component with the config. I rather incline to separate components. Not sure... What do you think?

pages/plugins.css
Joozty: I feel like flex is more configurable. Seems like block would be sufficient too.

pages/plugins.jsx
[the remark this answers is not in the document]
viktor-yakubiv: For the first time a generic component fits our needs. We cannot develop as much as in the design so we will have only installation instructions inside each plugin, and basically only 2 plugins.

On the other hand, you have a good point. Plugins have different configuration options and installation steps.

pages/plugins.jsx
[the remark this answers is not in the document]
viktor-yakubiv: Why do you actually need this?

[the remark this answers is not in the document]
Joozty: 🤷‍♂

pages/_app.jsx
Joozty: Maybe but not at this time.

Joozty:
@viktor-yakubiv It's based on several boilerplates including mentioned.

I don't really know how mobX works under the hood. I mean I know there are some getters/setters listeners but that's all. I don't know how would I do it. I just think it's necessary to have it. As I wrote it commit message. It can bring bugs into the code.

By the way, when I create a store with `Object.preventExtensions` does the mobX create a deep copy of that object instead of using it? It seems like that. Otherwise, it should work.

viktor-yakubiv:
One way to develop own MobX is just to use [Proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy). Another way is to deeply clone objects and return something like a proxy. They went second path.

So yes, from the [documentation](https://mobx.js.org/refguide/object.html) you may found that it's a deep copy:

> `observable` is by default applied recursively, so if one of the encountered values is an object or array, that value will be passed through observable as well.

> `observable` is applied recursively to a whole object graph automatically. Both on instantiation and to any new values that will be assigned to observable properties in the future. Observable will not recurse into non-plain objects.

pages/_app.jsx
viktor-yakubiv: It will make you more changes in the future. It's up to you do decide. I would prefer write everything properly form the start if I know about it know.

I am not a big fan of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) in such cases.

PR #19, opened 2019-12-09
Adjustments

Joozty:
@viktor-yakubiv This commit https://github.com/oacore/dashboard/pull/19/commits/63e82ceaff7a37e15e8bb74c495d52332932d44c maybe be particularly most important one. The svg sprite is copied to public folder after installation. If you have a better solution please share it with me.


components/layout/index.css
```diff
@@ -23,7 +23,7 @@
 }
 
 .layout-bar {
-  height: 3.5rem;
+  height: 4rem;
```
viktor-yakubiv: I wander, why this did change?

components/sidebar/index.jsx
```diff
@@ -53,8 +53,8 @@ const Sidebar = React.memo(({ className }) => {
             <Link href={route.path}>
               <a className={sidebarClassNames.route} href="/">
                 <Icon
-                  className={sidebarClassNames.sidebarIcon}
-                  iconType={route.icon}
+                  alt={`${route.title} icon`}
+                  src={`/design/icons.svg#${route.icon}`}
```
viktor-yakubiv: Maybe create a utility function for this or another component that does this automatically. Also, I am considering a CDN solution

next.config.js
```diff
@@ -60,13 +60,13 @@ const nextConfig = {
 
     config.module.rules.push({
       test: /\.css$/,
-      exclude: /@oacore\/design\/lib/,
+      exclude: /\/design\/lib/,
```
viktor-yakubiv: What does this mean?

package.json
```diff
@@ -20,6 +20,7 @@
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-prettier": "^3.1.0",
     "eslint-plugin-react": "^7.14.3",
+    "eslint-plugin-react-hooks": "^2.3.0",
```
viktor-yakubiv: We have to add it to global config. Feel free to make a PR

pages/plugins.jsx
```diff
@@ -33,23 +21,31 @@ const integrations = [
...
-          />
-        ))}
-      </div>
-    </>
-  )
+class Plugins extends React.Component {
+  static async getInitialProps({ initialStoreData }) {
+    const { statusCode, ...plugins } = await API.getIntegrations(140)
+    initialStoreData.plugins = plugins
+  }
```
viktor-yakubiv: Yeah, we can start from something like this. However, pages are presentation so we will need to move it somewhere unless I change my mind 😅

pages/plugins.jsx
```diff
@@ -33,23 +21,31 @@ const integrations = [
...
+
+  // eslint-disable-next-line class-methods-use-this
```
viktor-yakubiv: You can disable this globally if you experience this problem again.

Maybe an overriding rule for `pages/*.jsx`

components/searchbar/index.jsx
```diff
@@ -7,7 +7,7 @@ const SearchBar = () => {
     <div className={searchbarClassNames.searchbar}>
       <input
         type="text"
-        className={searchbarClassNames.searchbarInput}
+        value="Open Research Online - The Open University"
```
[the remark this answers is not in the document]
Joozty: We will need to discuss this more. It's not a priority now. I will take care of it after https://github.com/oacore/dashboard/pull/20

components/sidebar/index.css
```diff
@@ -59,8 +59,9 @@
   padding-bottom: 1rem;
 }
 
-.sidebar-navigation li .sidebar-icon {
-  width: 30px;
+.sidebar-navigation li svg {
+  width: 2rem;
+  height: 2rem;
```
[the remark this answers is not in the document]
Joozty: Is it too much? We will need to slightly rewrite the icon component to allow pass custom classNames and allow to wrap `svg` into another element.

components/layout/index.css
Joozty: Because of the input element height.

package.json
Joozty: You mean like a peerDependency or what?

components/sidebar/index.jsx
Joozty: Definitely I support it. CDN would solve this issue https://github.com/oacore/dashboard/pull/19#discussion_r356657904 too.

next.config.js
Joozty: We need to load CSS files from @design/library globally. Unfortunately, postcss doesn't support something like this:

```
:global {
	import url('@oacore/design/lib/index.css')
}
```

Therefore I needed to do it on webpack level. But there was a problem when @oacore/design was symlinked because regular expression didn't match.  Does it make sense?

pages/plugins.jsx
Joozty: I was considering it but I decided to give it another whirl.

pages/plugins.jsx
Joozty: Yeah, but everyone knows it's under development. I don't see any reason for hiding it. 😄

viktor-yakubiv: Architecture should be considered from the beginning. In this case this is not a big deal but we have to sort it out somehow later.

package.json
viktor-yakubiv: I mean React hooks support to global configuration if it's not there.

components/sidebar/index.css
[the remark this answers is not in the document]
viktor-yakubiv: Why do we need to wrap SVG?

components/layout/index.css
viktor-yakubiv: Input element height should be adjusted. You may use `size=small`

Joozty: Small didn't fit, i.e. it didn't look very well.

package.json
Joozty: It's inherited from airbnb-react, we don't need to configure anything for the time being.

next.config.js
viktor-yakubiv: I will think about it

PR #26, opened 2019-12-17
Use @oacore/design from GitHub

viktor-yakubiv:
This should speed up the development. However, something is going bad here.


viktor-yakubiv:
@Joozty if you can solve this one it would be great. Thank you!

viktor-yakubiv:
@Joozty thanks for your help. Closing in favour of #27

PR #28, opened 2019-12-18
Restructure application

viktor-yakubiv:
**Splits up Layout presentation and business logic**

1. Introduces new Application component for the business logic.
2. Keeps presentation logic, i.e. sidebar toggling, in the Layout component.
3. Splits Layout component to few more simple components.


viktor-yakubiv:
It's incredibly hard to work on good software architecture 😤

viktor-yakubiv:
This PR depends on #32. I suggest to review and merge #32 first, and then move back to this one to make the reviewing easier.

components/application/activity-select.jsx
```diff
@@ -0,0 +1,54 @@
...
+    title: 'Plugins',
+  },
+  {
+    path: '/settings',
+    icon: 'settings',
+    title: 'Settings',
+  },
+]
+
+const ActivitySelect = React.memo(({ className }) => (
```
Joozty: For me, it's just a primary/secondary menu... I think `ActivitySelect` is not a good name for this component.

components/layout/main.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { classNames } from '@oacore/design/lib/utils'
+
+import styles from './styles.css'
+
+const Main = ({ children, className, tag: Tag = 'main', ...restProps }) => (
```
Joozty: You kind of like to pass this `tag: Tag = '...'` everywhere. I don't know if it is a good idea. I see meaning in some cases. But for example, here it's completely useless. This component is used only once and I doubt it will be reused.

components/layout/side-bar.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import { AppBar } from '@oacore/design'
+import { classNames } from '@oacore/design/lib/utils'
+
+import { Consumer } from './context'
```
Joozty: Maybe `withGlobalStore` would be better.

components/layout/styles.css
```diff
@@ -0,0 +1,98 @@
...
+  padding: var(--component-padding, 1rem);
+  padding-top: calc(var(--app-bar-skip) + var(--component-padding, 1rem));
+
+  @media (--screen-until-desktop) {
+    z-index: 1050;
+    width: calc(100% - 3.5rem);
+    background: var(--white);
+    box-shadow: 0 2px 4px rgba(0, 0, 0, 0.25);
+    transition: transform 200ms ease;
+    transform: none;
```
Joozty: Is it necessary?

```suggestion
```

components/application/index.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+const Application = ({ children, ...restProps }) => (
+  <>
+    <Container {...restProps}>
+      <AppBar>
+        <RepositorySelect />
+      </AppBar>
+
+      <SideBar>
+        <h2>Navigate your data</h2>
```
Joozty: Is it going to stay here?

components/layout/app-bar.jsx
```diff
@@ -0,0 +1,32 @@
...
+        <AppBarToggle
+          className={styles['app-bar-toggle']}
+          type="button"
+          aria-haspopup="true"
+          aria-controls={sidebarId}
+          aria-expanded={sidebarOpen}
+          onClick={toggleSidebar}
+        />
+
+        <AppBar.Brand className={styles['app-bar-brand']} href="/">
```
Joozty: Everywhere else you use styles.element. It would be nice to stick with it. 
```suggestion
        <AppBar.Brand className={styles.appBarBrand} href="/">
```

components/application/index.jsx
viktor-yakubiv: Should be for screen reader only. I did not have this utility class for this moment.

components/layout/app-bar.jsx
viktor-yakubiv: How do you do suggested change? 🤔

components/layout/side-bar.jsx
viktor-yakubiv: I do not get it. Is it React API function?

In this case this is not about global. ID and state passing is only implementation details for accessibility reason. It should not be part of any external API. I used context to property hide such detail.

components/layout/main.jsx
viktor-yakubiv: It is more about consistency and clear implicit API. I prefer to allow passing rest props to actual DOM and changing node name. This is mostly for implementing accessibility too.

viktor-yakubiv: Also, I am considering reuse the element in 'activities' or actual pages to avoid using fragments.

Joozty: > This is mostly for implementing accessibility too.

I don't agree. If we do it like this everywhere the a11y eslint warnings/errors will never occur. Unfortunately, eslint is not that smart and cannot detect which element is representing Tag.

components/layout/side-bar.jsx
Joozty: Sorry, I thought you were exporting it from `/store` directory. It works like HOC.

https://github.com/oacore/dashboard/blob/master/store/index.jsx#L16
https://github.com/oacore/dashboard/blob/master/store/index.jsx

Anyway, do you really want to create multiple contexts/stores in the application? I have very very very bad experience to doing so. (hard to debug, hard to understand the code...) I would rather split the global store into multiple modules.

components/layout/app-bar.jsx
Joozty: It's pretty easy. Unfortunately, it works only for one line of code and it doesn't work in commits. You need to have open [all files](https://github.com/oacore/dashboard/pull/28/files).

https://twitter.com/github/status/1053360435074949120

components/layout/main.jsx
viktor-yakubiv: Good point 🤔

components/layout/side-bar.jsx
viktor-yakubiv: I do. I am just playing with technologies in this point. I do not really like React approach for such kind of things.

components/application/activity-select.jsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
+import Link from 'next/link' // TODO: Should be avoided
```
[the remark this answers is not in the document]
viktor-yakubiv: It ties us to Next.js. Even if we try to reimplement Link to avoid string dependency on Next, it makes us to import Link everywhere.

I would prefer to do it implicitly.

components/application/activity-select.jsx
viktor-yakubiv: Yes, it is. However, we should be more specific in naming to make clear what the product is.

components/layout/side-bar.jsx
Joozty: Then tell me why we needed MobX when you don't want to use it? We can have extra context for every component and not care at all about global state. I just have a bad feeling about it but not forcing you to change it. It's up to you.

components/layout/app-bar.jsx
viktor-yakubiv: This is cool anyway! Thank you!

components/application/index.jsx
Joozty: https://github.com/oacore/design/issues/61

components/application/activity-select.jsx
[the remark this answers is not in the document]
Joozty: I don't know. What's wrong with tieing us to Next.js? It was decided that we use Next.js. It's like we use bootstrap but we cannot import from it anything. Maybe I just don't understand. 

Link is a powerful component. It prefetches pages and does transition. Therefore I would not avoid it.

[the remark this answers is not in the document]
viktor-yakubiv: This is a good point. I will consider it. Thank you!

PR #29, opened 2019-12-18
Alias for design directory

Joozty:


next.config.js
```diff
@@ -74,6 +74,8 @@ const nextConfig = {
         postcssLoader,
       ].filter(Boolean),
     })
+
+    config.resolve.alias.design = `${__dirname}/design`
```
viktor-yakubiv: Thank you. I will keep this 2 lines only because the rest breaks my PR, if you do not mind.

Also, I am not sure, ESLint works with this. There should be eslint-plugin-import-alias added, or WebPack resolver for eslint configured.

viktor-yakubiv:
I am still wondering why ESLint passes

PR #32, opened 2019-12-19
Add a few minor enhancements

viktor-yakubiv:
1. Makes Node version requirement more tolerant
2. Adds `export Component from './component'` support
3. Add a few components to design

Is a part of #28.

### DIFF
--- a/design/app-bar/toggle.css
+++ b/design/app-bar/toggle.css
@@ -1,3 +1,5 @@
+@custom-media --screen-desktop (min-width: 728px);
+
 .container {
   color: var(--gray-700);
 }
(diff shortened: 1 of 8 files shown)


design/app-bar/toggle.css
```diff
@@ -0,0 +1,14 @@
+@custom-media --screen-desktop (min-width: 728px);
```
Joozty: I quite don't understand why you removed the centralized file for custom media. Now this custom media is repeated in several files...

viktor-yakubiv: Oh, thank you for pointing this. I have to remove it from here too.

viktor-yakubiv: Ideally this custom media should leave in a single file. I achieved this in the #28

viktor-yakubiv:
Thank you for the review!

PR #35, opened 2019-12-20
Add basic Card component

viktor-yakubiv:
I developed basic card component but I have not tested it. @Joozty what is the best place to integrate it?

### DIFF
--- a/design/card/index.css
+++ b/design/card/index.css
@@ -1,0 +1,15 @@
+:root {
+  --card-padding-x: var(--component-padding-x, 1rem);
+  --card-padding-y: var(--component-padding-y, 1rem);
+  --card-background: var(--white, #fff);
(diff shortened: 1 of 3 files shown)


Joozty:
I guess you can test it [here](https://github.com/oacore/dashboard/blob/master/pages/index.jsx#L7) with some mocked data.

viktor-yakubiv:
The Overview page is not going to be a part of the MVP so I would not like to spend any time on it. I will find something and let you know. Thank you!

Joozty:
@viktor-yakubiv Ok then. Please rebase and I think it can be merged.

Joozty:
Card tested here https://dashboard-git-deposit-dates.oacore.now.sh/deposit-dates

PR #37, opened 2019-12-20
Fix Head component

viktor-yakubiv:
Moves Head to Application — the only place it's used (for
this moment).

Removes extra styles.

Fixes spacing.


viktor-yakubiv:
By the way, I just discovered that [it's not good idea to call a branch `head`](https://stackoverflow.com/a/1692926/7205866).

PR #38, opened 2019-12-20
Improve sidebar look

viktor-yakubiv:
It styles secondary navigation at the sidebar. Drawer's interface is not really good, I hope we will improve it later. Also, I would appreciate any feedback about naming, I borrowed my one from [Material Design](https://material.io/components/navigation-drawer/).

It breaks navigation transitions because of Link removing. I will fix it in separate PR.

### DIFF
--- a/components/application/activity-select.css
+++ b/components/application/activity-select.css
@@ -1,32 +1,21 @@
-.sidebar-navigation,
-.sidebar-navigation li > .route {
-  padding: 0;
-  margin: 0;
-  text-indent: 0;
-  list-style-type: none;
+/* Should be deleted after fix in Icon component */
+.item-icon {
+  display: inline-block;
+  width: 1.5rem;
+  height: 1.5rem;
+  vertical-align: middle;
+  fill: currentColor;
 }
 
-.sidebar-navigation li > .route {
-  display: flex;
-  align-items: center;
-  color: inherit;
-  text-decoration: none;
-}
+/* stylelint-disable-next-line no-duplicate-selectors */
+.item-icon {
+  width: 1.2rem;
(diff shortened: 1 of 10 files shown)


viktor-yakubiv:
@Joozty this is ready for the review,

design/drawer/item.jsx
```diff
@@ -0,0 +1,23 @@
...
+  </Tag>
+)
+
+const TreeItem = ({ children, ...passProps }) => (
+  <li>
+    <FlatItem {...passProps}>{children}</FlatItem>
+  </li>
+)
+
+const Item = ({ children, tag = 'li', ...passProps }) => {
```
Joozty: Interesting sometimes you call it passProps and sometimes restProps? Is there any difference?

viktor-yakubiv: 'rest' if I actually destructurise, pass if I just forward without destructuring.

This particular case is special, I destrucutrise to make correct forwarding.

viktor-yakubiv: Maybe I should always call it `forwardProps`. What do you think?

components/application/activity-select.css
```diff
@@ -1,32 +1,21 @@
-.sidebar-navigation,
-.sidebar-navigation li > .route {
-  padding: 0;
-  margin: 0;
-  text-indent: 0;
-  list-style-type: none;
+/* Should be deleted after fix in Icon component */
```
Joozty: Would be nice to create task for it. :)

design/drawer/item.jsx
Joozty: forwardProps sounds better but passProps is good too

PR #39, opened 2019-12-20
Deposit dates

Joozty:


viktor-yakubiv:
The branch seems to be inappropriately rebased.

pages/deposit-dates.css
```diff
@@ -0,0 +1,28 @@
+.table thead > :first-child {
+  border-top-width: 2px;
+}
+
+.container {
+  display: grid;
```
viktor-yakubiv: It's better to use flex, in this case and mostly. Check it for mobile.

With flex you can provide grow and shrink properties to make it resize automatically screens size independently. With greed you need a media query.

Add please 18rem minimal width to card component. It should fix the problem here if you use flex.

viktor-yakubiv:
> The branch seems to be inappropriately rebased.

Sorry, this is my fault. I merged #35. You can b are on `master` now.

pages/deposit-dates.css
Joozty: Fair enough, thanks.

components/table/index.jsx
```diff
@@ -69,7 +77,7 @@ class InfiniteTable extends React.Component {
             value={searchTerm}
           />
         )}
-        <Table>
+        <Table className={className}>
```
viktor-yakubiv: You do not want to pass it using rest props 🙂

pages/deposit-dates.css
```diff
@@ -0,0 +1,49 @@
+/* TODO: Custom media from layout.css doesn't work */
+@custom-media --screen-until-desktop (max-width: 727px);
+
+.table thead > :first-child {
```
viktor-yakubiv: Okay, I will note this

pages/deposit-dates.css
```diff
@@ -0,0 +1,49 @@
...
+  border-top-width: 2px;
+}
+
+.row {
```
viktor-yakubiv: I expected to have a single flex container that is actually main. Do you mind if I play with styles a little bit more?

components/table/index.jsx
Joozty: true, thanks

pages/deposit-dates.css
Joozty: Go ahead. You are a CSS Pro. 🙂

Joozty:
@viktor-yakubiv Thank you for review even during the weekend. 🙂

pages/deposit-dates.css
viktor-yakubiv: Check #41 🙂

PR #40, opened 2019-12-21
Add navigation map

viktor-yakubiv:
I am trying to split presentational and business logic up but it's not the best option here. I will work more on it.

It's not finished yet, I am going to restructure URLs. However, it is ready for feedback. 9b86b27 is what I am basically proposing. I know it does not do prefetch but we can add it in our own if we need to.


components/application/navigation.jsx
```diff
@@ -0,0 +1,42 @@
...
+const titleMap = {
+  overview: 'Overview',
+  data: 'Data',
+  statistics: 'Statistics',
+  plugins: 'Plugins',
+  settings: 'Settings',
+}
+
+const captions = Object.fromEntries(
+  ['overview', 'data', 'statistics', 'plugins', 'settings'].map(activity => [
```
Joozty: This is duplicated in many places. I know still work in progress and you will probably fix it in next commits. The reason why I am pointing this out is that maybe we could consider having some config directory where all configs (tables, routes, plugins...) would be stored. What do you think?

components/application/navigation.jsx
```diff
@@ -0,0 +1,42 @@
...
+    activity,
+    <>
+      <Icon
+        className={styles.itemIcon}
+        src={`/design/icons.svg#${activity}`}
+        alt
```
Joozty: Is it good to pass empty alt?

viktor-yakubiv: Yes if you do not want text to be read (I guess).

components/application/navigation.jsx
viktor-yakubiv: Config directory is good idea. However, I am trying to avoid it as much as possible because once use have it, all your components become dependent on configs and things become broken.

It's duplicated for a reason – SRP:
- menu captions can be changed by a content manager or a designer
- urls are controlled by a developer

Here this is a shorthand mess to make it work. Captions should go somewhere to `@oacore/texts`. I am considering creating something in the dashboard for the time being and then move to separate package/repository when needed.

components/application/navigation.jsx
Joozty: Then why passing alt at all?

components/application/navigation.jsx
Joozty: Fair enough... Btw. what does SRP mean? 🤔

viktor-yakubiv: It's _Single Responsibility Principle_, the first one from SOLID. It means that only one actor should control a piece of code. If there exists a feature or flow, the only one person must cause a change. Even if workflow is the same for 2 or more actors for the current moment, it should be split up because in the future these actors may have different requirements.

components/application/navigation.jsx
viktor-yakubiv: `alt` is required by the [standard](https://html.spec.whatwg.org/multipage/images.html#alt).

In addition I found that it should not be empty but can be in such cases. In this case I do not want it to be read, so I pass `aria-hidden` below and I do not care if the image is not displayed because the button has a label for it. It's only styling. To make it properly better to use pseudo-elements but not for this time.

viktor-yakubiv:
This is what I wanted to achieve. Basically, I want to encode application state to URL as much as possible. Here I pass data provider ID to the URL.

In the future we can put more things, e.g. `offset` for the tables.

For time being the solution is not very sophisticated, I do not like it, so I would be happy to hear any architectural ideas from you @Joozty.

viktor-yakubiv:
This chains to #45. I will try to properly re-base it and prettify

viktor-yakubiv:
It becomes work in progress again

PR #41, opened 2019-12-21
Add deposit dates page styles

viktor-yakubiv:
@Joozty this would work. It breaks other pages look because we need use Main or Card instead Fragment. But we can fix it later when necessary.

### DIFF
--- a/pages/deposit-dates.css
+++ b/pages/deposit-dates.css
@@ -1,46 +1,47 @@
-/* TODO: Custom media from layout.css doesn't work */
-@custom-media --screen-until-desktop (max-width: 727px);
+%item {
+  flex: 3 3 90%;
+  margin: 0 var(--component-spacer-x, 1rem) var(--component-spacer-y, 1.5rem);
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  align-content: flex-start;
+  align-items: flex-start;
+  padding-right: 0;
+  padding-left: 0;
+
+  & > * {
+    @extend %item;
+  }
+}
 
 .table thead > :first-child {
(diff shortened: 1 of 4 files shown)


Joozty:
It looks ugly on page load. Is it related to the stuff you mentioned in the PR description?

https://www.loom.com/share/a8f382767ec04e00976fa20bf757085c

viktor-yakubiv:
Yes, it does. I thought it only on my computer something. I do not why it distributes the space if containers are small.

I will take a look at it.

pages/deposit-dates.css
```diff
@@ -1,46 +1,47 @@
...
-.row {
+.container {
   display: flex;
-  flex-wrap: nowrap;
-  margin-bottom: 2rem;
-
-  @media (--screen-until-desktop) {
-    flex-wrap: wrap;
+  flex-wrap: wrap;
+  align-content: flex-start;
```
viktor-yakubiv: This line fixes [the bug](https://github.com/oacore/dashboard/pull/41#issuecomment-568205794).

pages/deposit-dates.css
```diff
@@ -1,46 +1,47 @@
-/* TODO: Custom media from layout.css doesn't work */
-@custom-media --screen-until-desktop (max-width: 727px);
-
-.table thead > :first-child {
-  border-top-width: 2px;
+%item {
+  flex: 3 3 90%;
+  margin: 0 var(--component-spacer-x, 1rem) var(--component-spacer-y, 1.5rem);
```
Joozty: Does this variable even exist? Can you please always explicitly add TODO comment when something should be created.

PR #44, opened 2019-12-23
Add @oacore/texts package emulation

viktor-yakubiv:
Adds `@oacore/texts` package


webpack/.eslintrc.yml
```diff
@@ -0,0 +1,6 @@
+env:
+  browser: false
+  node: true
+
+rules:
+  import/no-extraneous-dependencies: [error, { devDependencies: true }]
```
Joozty: Does this file inherit rules from parent eslint config automatically?

viktor-yakubiv: Yes. This is the idea of ESLint. All nested configs inherit and partially override rules.

PR #45, opened 2019-12-24
Alias most often used modules, cut the names

viktor-yakubiv:
The idea behind is to simplify writing for this particular project and simplify long workgflow:

```js
import { navigation } from '@oacore/texts/dashboard'

import { AppBarToggle } from 'design'
```

to **short one**:

```js
import { AppBar, AppBarToggle } from 'design'
import { navigation } from 'texts'
```

---

On top of #44

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -7,8 +7,6 @@ env:
 
 settings:
   import/resolver:
-    alias:
-      - ['@oacore/texts/dashboard', './texts']
     node:
       paths: ['.']
       extensions: ['.js', '.jsx']
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -6,6 +6,10 @@ env:
   browser: true
 
 settings:
+  import/ignore:
+    # Ignores blind export of everything from node_modules as well as it's done
+    # by default
+    - design/index.js$
   import/resolver:
     node:
       paths: ['.']
--- a/design/index.js
+++ b/design/index.js
@@ -1,3 +1,5 @@
+export * from '@oacore/design/lib'
+
 export AppBarToggle from './app-bar/toggle'
 export Avatar from './avatar'
 export Card from './card'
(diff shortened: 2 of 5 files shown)


.eslintrc.yml
```diff
@@ -6,6 +6,10 @@ env:
   browser: true
 
 settings:
+  import/ignore:
+    # Ignores blind export of everything from node_modules as well as it's done
+    # by default
+    - design/index.js$
```
Joozty: I don't understand why you added this here. I read about the [rule](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md#ignore) but it didn't make any sense to me why it had to be included here. Did you experience any issues with importing `desing`?

viktor-yakubiv: Yes, I experienced issues with it. It ignores `node_modules`. So, if I import with wildcard (`*`) from `node_modules`, the linter cannot resolve anything and throws an error.

viktor-yakubiv:
It's more like proposal. To my mind is good to _write less, do more_ 🙂

.eslintrc.yml
Joozty: I see... Nonetheless, I tried it without it and it didn't throw any error (it's the default behaviour as you wrote in a comment, isn't it?). I guess these lines are not necessary, are they?

viktor-yakubiv: It does not throw because in this PR contains no import from `design`

Joozty: What about this https://github.com/oacore/dashboard/blob/alias/design/index.js#L1?

I also tried to change it like this but again no error was thrown:
```
export * from '@oacore/design'
import * as D from '@oacore/design'
```

viktor-yakubiv: Try to merge [this}(https://github.com/oacore/dashboard/blob/alias/components/layout/app-bar.jsx#L2) with line 7 in the same file and test again.

Joozty: Ohh I get it now. Thanks. It is a shame it doesn't work very well (IDE shows me a warning). This is another win for typescript. 😄

PR #57, opened 2020-01-02
Include data-provider ID to URL

viktor-yakubiv:
Some magic. More explanation in the commits. The main idea is to reflect application state to URL as much as possible.

On top of #58.


pages/data-providers/[data-provider-id]/index.js
```diff
@@ -0,0 +1,18 @@
...
+const Index = () => {
+  const router = useRouter()
+
+  useEffect(() => {
+    const pathname = `${router.pathname}/overview`
+    const slash = router.asPath.substr(-1) !== '/' ? '/' : ''
+    const asPath = `${router.asPath}${slash}overview`
+
+    router.replace(pathname, asPath)
+  })
```
Comment (Joozty) [inline pages/data-providers/[data-provider-id]/index.js]: You should pass empty array as a second parameter to match `componentDidMount` method.

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/index.js]: To useEffect, you mean?

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/index.js]: yes

components/application/route.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Route is developed to simplify constructing routes within the application.
+ * It uses class instead a function to be extendible.
```
Joozty:
```suggestion
 * It uses class instead of a function to be extendible.
```

pages/_app/route.js
```diff
@@ -0,0 +1,24 @@
+import Route from 'components/application/route'
+
+/**
+ * NextRoute is a Route specific for the Next.js. It defines `href` and `as`
+ * properties according to the `next/router` specification.
+ */
+class NextRoute extends Route {
```
Joozty: how is this used?

components/application/route.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  toString() {
+    return [`/data-providers/${this.dataProvider}`, this.activity]
+      .filter(notEmpty => notEmpty)
+      .join('/')
+  }
+}
+
+Route.parse = routeString => {
```
Joozty: I wonder why this cannot be static method inside Route?

viktor-yakubiv: That actually is. static keywords does the same I did here.

Here I use Route inside so I decided to write separate function to be clear that thinks I use here are declared above.

pages/_app/route.js
viktor-yakubiv: I App to handle `href` and `as` parameters for router

PR #61, opened 2020-01-03
Centralise texts, move captions to @oacore/texts

viktor-yakubiv:
On top of #60

### DIFF
--- a/texts/navigation.yml
+++ b/texts/navigation.yml
@@ -1,0 +1,10 @@
+title: Navigation
+items:
+  overview: Overview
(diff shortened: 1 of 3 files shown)


texts/navigation.yml
```diff
@@ -0,0 +1,10 @@
+title: Navigation
```
Joozty: This is not used you can remove it, can't you?

viktor-yakubiv: It's not used yet. It should be screen-reader only heading. Thank you

PR #62, opened 2020-01-03
Display current active tab

viktor-yakubiv:
On top of #61


pages/_app/index.jsx
```diff
@@ -32,23 +20,48 @@ class App extends NextApp {
         (await Component.getInitialProps({ ...ctx, initialStoreData })) || {}
     }
 
+    const route = new Route(
+      (ctx.req && ctx.req.url) || window.location.pathname
```
Joozty: when `ctx.req` is empty?

viktor-yakubiv: Oh, I do not use it anymore! Thank you!

It's empty on client side and fulfilled on server side.

PR #64, opened 2020-01-06
Login

Joozty:
@viktor-yakubiv This may be hard to review so let me summarize what is included in this pull request.

1. If `/user` fails then user gets redirected to login form
2. If any request fail on `401`. User is immediately redirected to login too. (I need to test it once backend sends this code)
3. If user tries to access dataProvider which is not equal to organisationId. User gets redirected to correct organisationId
4. Repository select is disabled. 
5. User permissions are not included in this PR.(Once we get permission we will enable repository select for admins)

I had to replace axios with fetch. You know the reasons. It didn't work even after OPTIONS fix.

More about the issue:
https://github.com/axios/axios/issues/587
https://github.com/axios/axios/issues/1661


viktor-yakubiv:
Is this moving in any way?

Joozty:
@viktor-yakubiv I am waiting for CORS header to be enabled. Backend guys are notified and once it's done I will continue.

api/index.js
```diff
@@ -1,41 +1,45 @@
...
       params,
-      headers: {
-        'Content-Type': 'application/json',
-        ...headers,
-      },
+      credentials: 'include',
+      headers,
     })
+
+    if (response.status === 401) Router.push('/login.html')
```
viktor-yakubiv: This is unexpected behaviour. `apiRequest` should not do it by definition, only request. This should be handled in store.

api/index.js
```diff
@@ -1,41 +1,45 @@
...
+    if (response.status >= 400 && response.status < 599)
```
viktor-yakubiv: Technically, either `< 600` or `<= 599`. Practically, `>= 400` is enough because there is no codes after _5XX_

components/application/index.jsx
```diff
@@ -11,9 +11,11 @@ const Application = ({ children, dataProvider, activity, ...restProps }) => (
...
     <Head />
     <Container {...restProps}>
-      <AppBar>
-        <RepositorySelect value={dataProvider} />
-      </AppBar>
+      {dataProvider && (
+        <AppBar>
+          <RepositorySelect value={dataProvider} />
+        </AppBar>
+      )}
```
viktor-yakubiv: You hide whole application bar if data provider is not set. It would be better to hide only RepositorySelect. In application bar can be other menu, user area etc.

components/application/repository-select.jsx
```diff
@@ -6,25 +6,18 @@ import styles from './repository-select.css'
...
       className={styles.repositorySelect}
-      options={fetchSuggestions}
-      value={store.repository.name}
+      options={store.dataProviders.retrieveDataProviders}
+      value={store.dataProviders.selectedProvider.name}
       onSelectionChange={v => {
         store.repository = v
         store.changeDataProvider(v.id)
       }}
+      disabled={store.user.isAdmin}
```
viktor-yakubiv: I am not sure about this but let's keep it for now.

design/select/index.jsx
```diff
@@ -24,6 +24,7 @@ const Select = ({
   value,
   options,
   className,
+  disabled = false,
```
viktor-yakubiv: ~It would not be necessary, if you would pass _htmlProps_.~ Sadly, it would.

pages/_app/index.jsx
```diff
@@ -21,12 +26,21 @@ class App extends NextApp {
...
 
-  store = null
-
-  componentDidMount() {
+  redirectToLogin = () => {
     const { router } = this.props
-    const { dataProvider, activity } = new Route(router.asPath)
-    this.store = initStore({ dataProvider, activity })
+    const pathname = '/login.html'
+    router.replace(pathname, pathname)
```
viktor-yakubiv: [Somewhere in the doc](https://nextjs.org/docs/api-reference/next/router#routerpush) I read:
> You don't need to use _Router_ for external URLs, `window.location` is better suited for those cases.

pages/_app/index.jsx
```diff
@@ -41,22 +55,29 @@ class App extends NextApp {
...
-  render() {
-    const { Component, pageProps, router } = this.props
-    const { store } = this
-
-    if (store === null) return null
+  static getDerivedStateFromProps(props, state) {
+    const { router } = props
+    const { store } = state
 
+    if (!store) return null
```
viktor-yakubiv: I am not sure, you can have `state == null`. You may find destructuring issues.

pages/index.js
```diff
@@ -1,16 +1,19 @@
...
+import { withGlobalStore } from 'store'
+
+const Index = ({ store }) => {
   const router = useRouter()
 
   useEffect(() => {
+    if (!store.user.organisationId) return
     const pathname = '/data-providers/[data-provider-id]'
-    const asPath = '/data-providers/1'
+    const asPath = `/data-providers/${store.user.organisationId}`
```
viktor-yakubiv: It's not an organisation. It should be actual data provider ID. The relationship is one organisation to many data providers.

public/login.html
```diff
@@ -60,14 +60,14 @@
 
 <script>
   function login(e) {
-    e.target.action = 'https://api.core.ac.uk/login_check?continue=' + encodeURI(
+    e.target.action = 'https://api.dev.core.ac.uk/login_check?continue=' + encodeURI(
```
viktor-yakubiv: It's a pain*. We will need to revert it back before release.

@mcancellieri can we deploy login to production and avoid this?

store/dataProviders.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  @action
+  retrieveDataProviders = async term => {
+    // TODO: Invalidate cache after some time
+    //       Move to @oacore/api
+    if (this.allProviders.get(term)) return this.allProviders.get(term)
+    const r = await fetch(
+      `https://api.core.ac.uk/internal/organisations?q=${term}`
+    )
```
viktor-yakubiv: It should be something like `https://api.core.ac.uk/internal/users/{id}/data-providers`

store/depositDates.js
```diff
@@ -120,7 +120,7 @@ class DepositDates {
       this.depositDatesCount = r.headers['collection-length']
     } catch (e) {
       if (e.statusCode === 404) this.isExportDisabled = true
-      else throw e
+      // else throw e
```
viktor-yakubiv: I guess this should not be commented. Up to you

store/index.jsx
```diff
@@ -9,13 +9,13 @@ let globalStore = null
...
-export const initStore = ({ dataProvider, activity }) => {
+export const initStore = () => {
   // on the server-side a new instance is created for each page request
   // as we don't wan't to mix between users/requests, etc.
-  if (isServer) return new RootStore(dataProvider, activity)
+  if (isServer) return new RootStore()
 
   if (!globalStore) {
-    globalStore = new RootStore(dataProvider, activity)
+    globalStore = new RootStore()
```
viktor-yakubiv: Do not you want to have some initial configuration? As a user you can go to or refresh `/data-providers/86/deposit-dates` and the application should load 86th data provider if the user has an access to it.

To extend, if user is not authorised, application should redirect to login form and after login it should redirect back to 86th data provider's deposit dates but we would not do it for now.

store/user.js
```diff
@@ -9,11 +15,35 @@ class User {
...
+    const user = await apiRequest('/user', 'GET', {}, {}, true)
+    this.rootStore.changeDataProvider(user.organisationId)
+    Object.assign(this, user)
+  }
+
+  @action
+  loadUserPermissions = async () => {
+    // if (!this.permissionsUrl) return
+    // const permission = await apiRequest(this.permissionsUrl)
+  }
```
viktor-yakubiv: I am not sure we need explicit user permissions for this moment but whatever you wish. If you want to go that way, try to localise permission tags as much as possible because it might be extremely hard to rename something at some point.

store/user.js
```diff
@@ -9,11 +15,35 @@ class User {
...
+  @action
+  logIn = async () => {
+    await this.retrieveUser()
+  }
+
+  @action
+  retrieveUser = async () => {
+    const user = await apiRequest('/user', 'GET', {}, {}, true)
+    this.rootStore.changeDataProvider(user.organisationId)
+    Object.assign(this, user)
```
viktor-yakubiv: I am considering this being bad pattern but I am too lazy to do in other way. So, 👍

store/store.js
```diff
@@ -45,9 +44,16 @@ class RootStore {
...
       if (isServer) return
-      const route = new Route(window.location.pathname)
-      route.dataProvider = this.dataProvider
-      Router.push(route.href, route.as)
+
+      // if user has no access to dataProvider redirect them
+      const route = new Route(Router.asPath)
+      if (
+        route.dataProvider &&
+        parseInt(route.dataProvider, 10) !== this.user.organisationId
```
viktor-yakubiv: One more line with comparison an organisation to a data provider

pages/index.js
viktor-yakubiv: To extend what I wrote here, there is a fresh [task](https://kmi-ou.atlassian.net/browse/CORE-2917) for @valerapro. Unfortunately, we are partially blocked.

pages/_app/index.jsx
Joozty: Interesting. Is `./login.html` considered as external url?

store/dataProviders.js
Joozty: Yep this was done long time ago. We need to change it but sadly it's out of scope for this PR.

pages/_app/index.jsx
Joozty: When it can happen?

pages/_app/index.jsx
viktor-yakubiv: It cannot be loaded by Next so it should be considered 'external', in my opinion.

store/user.js
Joozty: We need it for repository select, don't we?

public/login.html
Joozty: Well, cookie will be set for api.core.ac.uk and not for api.dev.core.ac.uk.

api/index.js
Joozty: Well, in that case, we need a new API layer in store or to do redirected for every API call separately. I will leave it as is for now.

store/user.js
viktor-yakubiv: I assumed that `/users/{id}/data-providers` returns the list of only data providers that the user can edit. So, in this case, we can implicitly check existence of an ID in this list.

pages/_app/index.jsx
viktor-yakubiv: It can happen when you blindly start destructuring state in `render()` like `const { store } = this.state` what you actually do. If you try `const store = null.store` you will face the problem.

Joozty: I mean when `state` can be null?

viktor-yakubiv: As far as I understand the method returns new state based on current props and current state. If you return null you may get state being null. I am not sure now.

store/index.jsx
Joozty: There is no need to initial configuration. These values are set here `https://github.com/oacore/dashboard/blob/login/pages/_app/index.jsx#L65`.


> To extend, if user is not authorised, application should redirect to login form and after login it should redirect back to 86th data provider's deposit dates but we would not do it for now.

We will need to store the URL somewhere in storage/cookie. Now the redirection is hardcoded. Let's do it in another PR.

store/store.js
Joozty: I don't get it.

store/index.jsx
viktor-yakubiv: I would rather pass it as a `continue` parameter, cached by JS and passed again.

pages/_app/index.jsx
Joozty: `getDerivedStateFromProps is invoked right before calling the render method, both on the initial mount and on subsequent updates. It should return an object to update the state, or null to update nothing.`

https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

store/store.js
viktor-yakubiv: You compare a data provider ID to an organisation ID. These entities are 2 different so comparison is wrong.

pages/index.js
Joozty: I did a mistake. I thought dataProvider is equal to organisationId

pages/_app/index.jsx
viktor-yakubiv: Thank you for the explanation 👍

store/user.js
Joozty: I did not know about it. My fault. I will make changes accordingly.

viktor-yakubiv: The endpoint does not work correctly. You will need to add a workaround before @valerapro makes changes. Here is what I got till this moment:

```js
import apiRequest from 'api'

const fetchSuggestions = async (userId, query) => {
  const { data } = await apiRequest(
    `/users/${userId}/data-providers`,
    'GET',
    {
      q: query,
    },
    {},
    true
  )

  return data.organisation.repositories
}
```

public/login.html
viktor-yakubiv: I know. I am pointing that at some point we will need to replace all occurs of `api.dev.` to `api.` to make the application production really what may (definitely will) be painful.

PR #66, opened 2020-01-07
Background blur on repository select open

Joozty:


components/application/index.jsx
```diff
@@ -11,9 +10,7 @@ const Application = ({ children, dataProvider, activity, ...restProps }) => (
   <>
     <Head />
     <Container {...restProps}>
-      <AppBar>
-        <RepositorySelect value={dataProvider} />
-      </AppBar>
+      <AppBar dataProvider={dataProvider} />
```
viktor-yakubiv: Why could not you keep this as it was?

viktor-yakubiv:
I approve 9446b28 👍. Feel free to cherry pick it directly to `master` and then rebase this PR. Thank you!

---

Do you have any ideas how to add global overlay? For modals and drop-downs.

components/application/index.jsx
Joozty: I needed component connected to `LayoutContext`. (I didn't want to connect another component into it.)

Joozty:
I don't know how to do a global overlay for dropdowns. I am worried it has to be done separately for every dropdown as it is done in this PR.


For global overlay for modals, there are several options. Nonetheless, the best option for us is to use `React.Portals`. Modal would be rendered as a direct child of body element through the portal...

```
<body>
	<div id="__next" />
	<div classname="overlay">
		<div>Modal content here</div>
	</div>
</body>
````

File: components/layout/container.jsx [hunk 2]
```diff
@@ -8,6 +8,8 @@ const reducer = (state, action) => {
   switch (action.type) {
     case 'toggle_sidebar':
       return { ...state, sidebarOpen: !state.sidebarOpen }
+    case 'toggle_repository_select':
+      return { ...state, repositorySelectOpen: action.value }
```
viktor-yakubiv: This should be encapsulated under `CustomSelect`.

Joozty:
I don't consider it important. I just saw it in Figma design and wanted to try to implement it. I do agree it would be great to have some global overlay instead but this would blur all elements. Some of them (like dropdown) show be unblur. Maybe negative `filter: blur(-4px` would work. I don't know. 


Let's not close it. I may find some time to try different approaches. (It's not a priority for me for now.)

PR #67, opened 2020-01-09
Connect tables into API

Joozty:


pages/_app/index.jsx
```diff
@@ -5,28 +5,25 @@ import { withRouter } from 'next/router'
...
 
     return {
       pageProps,
-      initialStoreData,
+      store,
     }
   }
 
-  handleNavigation = event => {
+  handleNavigation = async event => {
```
viktor-yakubiv: I did not know about. Sorry and thank you!

pages/_app/index.jsx
```diff
@@ -35,19 +32,24 @@ class App extends NextApp {
...
+    store.dataProvider = dataProvider
+    store.activity = activity
+
     return (
-      <GlobalProvider initialData={initialStoreData}>
+      <GlobalProvider store={store}>
         <Application
-          dataProvider={dataProvider}
-          activity={activity}
+          dataProvider={store.dataProvider}
```
viktor-yakubiv: I think, it needs only `dataProviderId`, that can be passed as `dataProvider.id`. Everything inside can be connected independently.

I try to pass primitives to props as much as possible because passing object to HTML attribute seems to be weird. It's only because of JSX notation, I feel good passing objects using `React.createElement(Component, props, children)`

pages/_app/index.jsx
```diff
@@ -35,19 +32,24 @@ class App extends NextApp {
...
+          activity={store.activity}
```
viktor-yakubiv: I wonder if this can be calculated somehow based on PageComponent but not for this moment.

store/user.js
```diff
@@ -0,0 +1,15 @@
+class User {
+  id = null
+
+  name = null
+
+  givenName = null
+
+  familyName = null
```
viktor-yakubiv: We do not have these 2 fields I guess

store/user.js
```diff
@@ -0,0 +1,15 @@
+class User {
```
viktor-yakubiv: I love that! 👍

pages/_app/index.jsx
Joozty: It is not needed. I forgot to remove it. I was just testing something.

store/user.js
Joozty: I took them from some example response. I will remove it.

pages/_app/index.jsx
Joozty: Ok let me change it in next PR

PR #70, opened 2020-01-14
Redirect user after login to correct page

Joozty:


Joozty:
@viktor-yakubiv I am not sure whether it `continue` should be GET or POST param.

public/login.html
```diff
@@ -68,7 +77,7 @@ <h1>Login</h1>
   <input id="password" type="password" placeholder="Enter Password" name="password" required>
 
   <button type="submit">Login</button>
-  <label>
-    <input type="checkbox" checked="checked" name="remember"> Remember me
+  <label style="display: none;">
+    <input type="checkbox" checked="checked" name="remember_me"> Remember me
```
viktor-yakubiv: It would be shorter to align it with other inputs without any label:

```<input type="hidden" name="remember_me" value="on">```

viktor-yakubiv:
`continue` parameter can be either in the URL or in the body if @valerapro deployed changes.

Your implementation is the best I think so let's keep using GET parameter.

PR #73, opened 2020-01-15
Deposit Time Lag Chart

Joozty:


components/time-lag-chart/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+        tickLine={false}
+        ticks={ticks}
+        tickFormatter={formatter}
+      />
+      <Tooltip />
+      <ReferenceLine y={0} stroke="#424242" />
+      <Bar dataKey="worksCount">
+        {data.map(entry => {
+          const color = entry.depositTimeLag <= 90 ? '#5A9216' : '#9E9E9E'
+          return <Cell key={entry.depositTimeLag} fill={color} />
```
viktor-yakubiv: Prefer className. It's a SVG element, it should support CSS, so you can style with variables.

components/time-lag-chart/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+    case 365 * 2:
+      return '2y'
+    case 365 * 3:
+      return '3y'
+    default:
+      return `${value}`
+  }
+}
+
+const ticks = [-365, -31, -7, 0, 7, 31, 90, 365]
```
viktor-yakubiv: I would prefer automatic logarithmic axis. Is it possible? How much effort does it require? Do we need D3 for it?

Joozty:
> We should find a way to aggregate data and merge values into something like past 7 (or 10) days, past month, past year. I would look for already developed logarithmic scale aggregation

Isn't this work for the backend?

components/time-lag-chart/index.jsx
Joozty: Good idea! Log scale for YAxis shouldn't be that hard to do. Recharts has [API](http://recharts.org/en-US/api/YAxis#scale) for it but it didn't work. I will need to raise an issue. Nonetheless, we can use D3 directly if it doesn't work.

I am not sure about Xaxis. The type of Xaxis is categorical, not numeric. I don't you can apply log scale on this, can you? We would need to develop our own scale function. Let's wait for the backend to see what data we will get. :)

viktor-yakubiv:
I would expect pure data from back-end. It's not that much. 

Aggregating data to display is more about view task. We can do it if we have all data but we cannot split data if we have it aggregated. Also, querying API to get new aggregation because screen size changed is not good idea at all.

Joozty:
Fair enough I am with you. In this case, I am not sure whether bar chart is the best type of graph for this. If so we need to write our own scale function based on the viewport width.

viktor-yakubiv:
Let's keep it simple unless we face actual problem 🙂

PR #74, opened 2020-01-15
Small table fixes

viktor-yakubiv:
@Joozty I am experiencing issues that are not expected. Do you have any ideas why count with auto width renders actual zero width when `table-layout: fixed`?

### DIFF
--- a/components/published-toggle/index.jsx
+++ b/components/published-toggle/index.jsx
@@ -3,7 +3,7 @@ import classNames from 'classnames'
 
 import styles from './index.css'
 
-const PublishedToggle = ({ defaultVisibility, isExpanded }) => {
+const PublishedToggle = ({ className, defaultVisibility, isExpanded }) => {
   const [isChecked, toggleChecked] = useState(defaultVisibility)
   let toggletext = ''
   if (isExpanded) {
@@ -29,7 +29,7 @@ const PublishedToggle = ({ defaultVisibility, isExpanded }) => {
         }}
         checked={isChecked}
       />
-      <div className={styles.toggleSwitch}>
+      <div className={classNames(styles.toggleSwitch, className)}>
         <span>{toggletext}</span>
       </div>
     </label>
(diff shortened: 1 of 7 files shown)


Joozty:
I ran into that issue too. I don't know why it happens. 😞

viktor-yakubiv:
Maybe we can merge first 3 commits and think about the last one that breaks things.

components/published-toggle/index.jsx
```diff
@@ -29,7 +29,7 @@ const PublishedToggle = ({ defaultVisibility, isExpanded }) => {
         }}
         checked={isChecked}
       />
-      <div className={styles.toggleSwitch}>
+      <div className={classNames(styles.toggleSwitch, className)}>
```
Joozty: I had a thought about this and maybe we should start using only classNames from @oacore/design

pages/data-providers/[data-provider-id]/deposit-dates.css
```diff
@@ -41,23 +34,44 @@
   margin-bottom: 0;
   overflow-x: auto;
 
-  & > table {
-    min-width: 40rem;
```
Comment (Joozty) [inline pages/data-providers/[data-provider-id]/deposit-dates.css]: I would keep `min-height` here. Otherwise, it causes a horizontal overflow on small screens.

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/deposit-dates.css]: You mean min-width?

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-dates.css]: yes sorry

components/published-toggle/index.jsx
viktor-yakubiv: I was too lazy to update it. If you want to I will do this change 🙂

Joozty: No need, I will do it for all occurrences in one of my commits.

viktor-yakubiv:
This is not much, really a little thing I can do. I wonder why we have unexpected behaviour and guess I have to take closer look at sanitize.css. I saved it to #76

Meanwhile, we have other problems to fix.

pages/data-providers/[data-provider-id]/deposit-dates.css
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/deposit-dates.css]: I kept it. Thank you!

PR #81, opened 2020-01-17
Overview

viktor-yakubiv:


design/index.js
```diff
@@ -6,6 +6,7 @@ export Card from './card'
 export CloseIcon from './close-icon'
 export Drawer from './drawer'
 export Logo from './logo'
+export Numeral from './numeral'
```
viktor-yakubiv: @Joozty do you have any better ides if Number is reserved?

Joozty: I don't think there is anything better for this. Maybe `Digit`?

design/numeral/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+  </Tag>
+)
+
+const Diff = ({ children, className, tag: Tag = 'span', ...restProps }) => (
+  <Tag className={classNames.use(styles.diff).join(className)} {...restProps}>
+    {children}
+  </Tag>
+)
+
+const Value = ({ children, className, tag: Tag = 'span', ...restProps }) => (
```
Joozty: I would make these all React.memo. It's a very small improvement but still

pages/data-providers/[data-provider-id]/overview.css
```diff
@@ -0,0 +1,61 @@
+.container {
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(16rem, auto));
```
Comment (Joozty) [inline pages/data-providers/[data-provider-id]/overview.css]: Can you fix this spacing?

padding

design/numeral/styles.css
```diff
@@ -0,0 +1,51 @@
...
+
+.diff {
+  vertical-align: top;
+}
+
+.diff-none {
+  display: none;
+}
+
+.diff-positive::before {
```
Joozty: these classes are not used

viktor-yakubiv: Yet, we do not need it for this moment because it's overkill for this task.

pages/data-providers/[data-provider-id]/overview.css
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/overview.css]: I wish I could. I do not know how to handle this with grid but do not want to go back to flex and use workarounds.

pages/data-providers/[data-provider-id]/overview.jsx
```diff
@@ -1,17 +1,128 @@
...
+        <RadialBarChart
+          innerRadius="10%"
+          outerRadius="80%"
+          data={data}
+          startAngle={225}
+          endAngle={-45}
+        >
+          <RadialBar minAngle={15} background clockWise dataKey="uv" />
+        </RadialBarChart>
+      </ResponsiveContainer>
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/overview.jsx]: @Joozty `RadialBarChart` does not work as well as I expected. Unfortunately I have to fall back to own component.

pages/data-providers/[data-provider-id]/overview.css
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/overview.css]: I think the only issue here is the wrong number:

```
    grid-template-columns: repeat(auto-fill, minmax(18rem, auto));
```

pages/data-providers/[data-provider-id]/overview.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/overview.jsx]: I would not give it up just like. Are you sure about this? What about other solutions like Piechart?

```
const DepositingCard = ({ title = 'Depositing', description, value }) => {
  const chartData = [
    { name: 'score', value: 680, fill: 'var(--primary)' },
    { name: 'max', value: 400, fill: 'var(--gray-100)' },
  ]
  return (
    <Card>
      <h2>{title}</h2>
      <ResponsiveContainer width="100%" height={300}>
        <PieChart margin={{ top: -25, right: -25, left: -25, bottom: 0 }}>
          <Pie
            data={chartData}
            cx="50%"
            cy="50%"
            startAngle={-135}
            endAngle={-405}
            innerRadius="60%"
            outerRadius="80%"
            stroke="none"
          />
        </PieChart>
      </ResponsiveContainer>
      <Num value={value} append="%" caption="papers" />
      <p>{description}</p>
      <Button variant="contained" href="deposit-dates" tag="a">
        Browse
      </Button>
    </Card>
  )
}
```

I did not style it properly. If you think it easier to use something else for you then go for it but I still think you should give it another try.

pages/data-providers/[data-provider-id]/overview.css
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/overview.css]: Card has minimal width itself what I would not like to change because it affects flex containers in other places.

There should be another solution but I do not know it yet. I would like to postpone this if you do not mind.

viktor-yakubiv:
@Joozty I finished basic layout. To my mind, that would be enough for this moment despite it has a few bugs. Could you please review?

design/numeral/styles.css
Joozty: Can you remove them then?

design/numeral/index.jsx
Joozty: Can you do it?

viktor-yakubiv: Sorry, yes

PR #82, opened 2020-01-20
Use `/public/release-dates` in deposit dates table

Joozty:


next.config.js
```diff
@@ -9,6 +9,15 @@ const nextConfig = {
   webpack(config, options) {
     const { dev, isServer } = options
 
+    const originalEntry = config.entry
+    config.entry = async () => {
+      const entries = await originalEntry()
+      if (entries['main.js'])
+        entries['main.js'].unshift('@oacore/design/lib/index.css')
```
viktor-yakubiv: Yes, this is the solution. It would be nice if you could write more verbose commit message to cover the bug. Thank you for doing this @Joozty!

Joozty: This was not a solution. This file needed to be imported in some js file.

I ended up creating two js/css chunks. One for node_modules and one for everything else. I wanted to create one chunk for `design` css and one for everything else. Since there was order problem again. (I think chunk order is based on the order of imports). Hence I grouped all `node_modules`.

PR #83, opened 2020-01-20
CSV Export

Joozty:
@viktor-yakubiv There is an issue with `Access-Control-Allow-Origin` so it's hard to test it. It worked for me a few minutes ago.

### DIFF
--- a/store/depositDates.js
+++ b/store/depositDates.js
@@ -1,4 +1,5 @@
-import { computed } from 'mobx'
+import download from 'downloadjs'
+import { action, computed, observable } from 'mobx'
 
 import Page from './helpers/page'
 import getOrder from './helpers/order'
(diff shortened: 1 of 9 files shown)


Joozty:
@viktor-yakubiv I was thinking about it too but IMHO it violated principle to have API calls only in store action. Also with this approach, we can show some error when download fails (API down or anything else)

viktor-yakubiv:
We can accept such behaviour for this moment but in the future we should display a modal window with request to buy the product.

store/depositDates.js
```diff
@@ -66,6 +80,55 @@ class DepositDates {
...
+          accept: 'text/csv',
+        },
+        {
+          'Content-Type': 'text/csv',
+        },
+        true
+      )
+      await download(data[0], 'deposit-dates.csv', 'text/csv')
+    } finally {
+      this.isExportInProgress = true
```
viktor-yakubiv: @Joozty is not this a mistake?

PR #85, opened 2020-01-21
Table sort functionality

Joozty:


components/table/index.jsx
```diff
@@ -148,6 +148,7 @@ class InfiniteTable extends React.Component {
               this.setState({
                 searchTerm: event.target.value,
                 page: 0,
+                isEmpty: false,
```
viktor-yakubiv: I there really no way to calculate this?

store/depositDates.js
```diff
@@ -11,21 +12,34 @@ class DepositDates {
...
+      from: pageNumber * PAGE_SIZE,
+      size: PAGE_SIZE,
+    }
+
+    if (order) params.orderBy = order
+    if (searchTerm) params.q = searchTerm
+
+    let data
+    try {
+      ;[data] = await apiRequest(
```
viktor-yakubiv: Why API request does return an array?

Joozty: Because I designed it that way. 😄 It was not a clever move. Anyway, I prepared a refactor of it...coming in following PR.

components/table/index.jsx
Joozty: Nope table component works like a proxy. It doesn't know about exact data in pages.

PR #86, opened 2020-01-21
Calculate compliance level

Joozty:
@viktor-yakubiv I hope formula is correct.


Joozty:
@viktor-yakubiv I don't think so. You need to iterate through the array anyway. I don't see any particular speedup in that. Or did you mean it for a different reason?

viktor-yakubiv:
I meant mostly speed up. However, if the number or algorithm changes it will affect all clients or a single API.

We should think more about it.

PR #87, opened 2020-01-21
Develop Plugins

WebDevelopUa:


Joozty:
@viktor-yakubiv I am interested in how this is gonna be styled. I guess there is no way to attach local class names in .md files.

Joozty:
@WebDevelopUa I think it needs some styling before the merge. Could you please do that?

viktor-yakubiv:
@Joozty what we have to solve somehow, where to store images for these instructions. Any ideas?

viktor-yakubiv:
@mcancellieri I will need your help with improving installation instructions texts. See example for [the Recommender instructions](https://dashboard-git-installation-instructions.oacore.now.sh/data-providers/86/plugins/recommender).

I already developed a functionality that replaces a `{{ key }}` with actual key. At this moment the `key` variable is `XXXXX` but we will improve it later.

mcancellieri:
What is that you need my help for?

viktor-yakubiv:
Could you please review [these texts](https://github.com/oacore/dashboard/pull/87/files?file-filters%5B%5D=.md) to be correct texts that are displayed directly in the dashboard. I will take care on images and styles but actual texts should refined according to the assumption that we injected the key directly into code snippets. Also, it would be nice to make instructions shorter if possible.

You can commit directly on GitHub.

texts/plugins/discovery.md
```diff
@@ -1,6 +1,136 @@
...
+```html
+<script
+  type="text/javascript"
+  src="https://discovery.core.ac.uk/plugin.js?template=eprints&amp;id={{ key }}" 
+  async="async"
+></script>
+```
+
+CORE&nbsp;Discovery repository plugin will automatically position itself after
+the heading and author section in the metadata page. However,  if you experience
```
Joozty:
```suggestion
the heading and author section in the metadata page. However, if you experience
```

texts/plugins/discovery.md
```diff
@@ -1,6 +1,136 @@
...
+any issues with automatic snippet injection, copy the following code and paste
+it wherever you wish the CORE&nbsp;Discovery to appear.
+
+```html
+<div id="core-discovery-root"></div>
+```
+
+You can take a glimpse of the&nbsp;plugin at  [Open Research
```
Joozty:
```suggestion
You can take a glimpse of the&nbsp;plugin at [Open Research
```

texts/plugins/discovery.md
```diff
@@ -1,6 +1,136 @@
...
+  </a>
+</div>
+
+To customise the&nbsp;plugin, go through the Custom styling section.
+
+
+## Everyone else (non-Eprints)
+
+Install the CORE&nbsp;Discovery repository plugin in any non-EPrints repository
+or website. Add the following snippet on every web page on the site you wish
```
Joozty:
```suggestion
or website. Add the following snippet on every web page you wish
```

texts/plugins/discovery.md
```diff
@@ -1,6 +1,136 @@
...
+
+Also, add a `div` element with the following structure wherever you wish
+the&nbsp;plugin to appear.
+
+```html
+<div id="core-discovery-root"></div>
+```
+
+After the installation, if the full text is not available from the repository
+but  is discovered by CORE, the script will display a link to the full text as
```
Joozty:
```suggestion
but is discovered by CORE, the script will display a link to the full text as
```

texts/plugins/discovery.md
```diff
@@ -1,6 +1,136 @@
...
+```css
+#core-discovery-root a {
+  text-decoration: underline;
+}
+```
+
+We recommend that you rely on element types (e.&nbsp;g.&nbsp;`table`, `a`, `p`) 
+or any other generic selectors (i.&nbsp;e.&nbsp;attribute, pseudo-classes)  and
+do not recommend to customise styles based on our class names like 
+`core_label__1ORqN` because they are generated automatically  and may change
```
Joozty:
```suggestion
`core_label__1ORqN` because they are generated automatically and may change
```

texts/plugins/discovery.md
```diff
@@ -1,6 +1,136 @@
...
+## How to blacklist a resource
+
+In case you would like to block the&nbsp;plugin for a specific resource,
+[email us][team-email] with the DOI of the resource.
+
+## Troubleshooting
+
+CORE&nbsp;Discovery repository plugin crawls the meta-tags of a metadata record
+in the repository to check the availability of full text. If the tag
+`citation_pdf_url`  does not exist the&nbsp;plugin is triggered, otherwise
```
Joozty:
```suggestion
`citation_pdf_url` does not exist the&nbsp;plugin is triggered, otherwise
```

components/application/route.js
```diff
@@ -30,9 +30,9 @@ Route.parse = routeString => {
   const s = routeString.charAt() === '/' ? routeString.slice(1) : routeString
 
   // Skip '/data-providers/' part
-  const [, dataProvider, activity] = s.split('/')
+  const [, dataProvider, ...activity] = s.split('/')
```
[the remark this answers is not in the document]
viktor-yakubiv: Thank you!

viktor-yakubiv:
It seems I am blocked by #148. I will extract some code into separate PR.

pages/data-providers/[data-provider-id]/plugins/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+  const [isLoading, setLoading] = useState(false)
+  const handleEnable = event => {
+    if (onEnable) {
+      onEnable(event)
+      setLoading(true)
+    }
+  }
+
+  // TODO: If there is a loading state we have to automatically perform
+  // route change after enabling
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/plugins/index.jsx]: @Joozty I am struggling to do this, and struggling much more to do this minimising side effects.

The experience should be the following:
1. A user visits `/plugins` page
2. There is a _Discovery_ plugin that is not enabled yet
3. The user presses _Enable_ and sees button changed state to loading
4. After successful enabling (generating a key) the user should see the instructions on the another page

Ideally the router call should be implicit but if it's not possible, explicit is fine too.

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/plugins/index.jsx]: This seems to be wrong. I will rethink the process.

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/plugins/index.jsx]: You are looking probably for something like this

import React, { useState, useEffect, useCallback } from 'react'

```
const PluginCard = ({
  title,
  description,
  enabled,
  onEnable,
  href,
  ...restProps
}) => {
  const [isLoading, setLoading] = useState(false)

  // memoize this function, i.e. don't recalculate it on every render
  const handleEnable = useCallback(async event => {
    if (onEnable) {
      try {
        await onEnable(event)
        setLoading(true)
      } catch (ignoreError) {
        // ignoring API error
      }

    }
  }, [])

  useEffect(() => {
    // false -> true
    if (isLoading) {
      // handle route change here
    }
  }, [isLoading]) // this function is called only when isLoading value changes

  return (
    ...
  )
}

```

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/plugins/index.jsx]: @mcancellieri this user experience is caused by back-end and it's not really good. I see the following options:
1. Generate the key implicitly when the user visits instructions page
2. Generate keys implicitly when the user visits generic plugins page
3. Generate keys implicitly once user registers in the Dashboard

Comment (mcancellieri) [reply pages/data-providers/[data-provider-id]/plugins/index.jsx]: yes that's fine. In the previous dashboard we generated the key as soon as the user opened the instructions tab... would that work?

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/plugins/index.jsx]: I you can generate the key first time `/plugins` endpoint is requested responding always with keys without affects on performance, it would be amazing.

@valerapro FYI

Comment (valerapro) [reply pages/data-providers/[data-provider-id]/plugins/index.jsx]: Yes, it possible generate key for first time request.

viktor-yakubiv:
@Joozty could you please review or approve this? I am afraid to merge the feature without review

components/application/route.js
```diff
@@ -30,7 +30,8 @@ Route.parse = routeString => {
   const s = routeString.charAt() === '/' ? routeString.slice(1) : routeString
 
   // Skip '/data-providers/' part
-  const [, dataProvider, activity] = s.split('/')
+  const [, dataProvider, ...activityPartials] = s.split('/')
+  const activity = activityPartials.join('/')
```
viktor-yakubiv: @Joozty by the way, `.split('/', 3)` does not work because of the following (from the [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split)):

> If provided, splits the string at each occurrence of the specified `separator`, but stops when `limit` entries have been placed in the array. **Any leftover text is not included in the array at all.**

PR #89, opened 2020-01-22
Connect real data to Overview

viktor-yakubiv:
@Joozty I would like to get your early feedback about connecting store. I am afraid I miss calling events somewhere and do not handle async methods. Could you please point me where to look at and what is the best way to fetch statistics data if it is needed only once at this moment?

On top of #81. Please, check the last commit only.

---

**UPD:** This PR should have been a draft but I have mistaken.

### DIFF
--- a/store/statistics.js
+++ b/store/statistics.js
@@ -1,0 +1,43 @@
+import { action, extendObservable } from 'mobx'
+
+import apiRequest from 'api'
+
+const initialStore = {
+  metadataCount: 0,
+  fullTextCount: 0,
+  doiCount: 0,
+}
+
+class Statistics {
+  constructor(rootStore) {
+    this.rootStore = rootStore
+    extendObservable(this, initialStore)
+  }
+
+  retrieveStatistics = async (dataProviderId = this.rootStore.dataProvider) => {
+    try {
+      const { data } = await apiRequest(
+        `/data-providers/${dataProviderId}/statistics`,
+        'GET',
+        {},
+        null,
+        true
+      )
+
+      return data || {}
+    } catch (e) {
+      if (e.statusCode !== 404) throw e
+      return {}
+    }
+  }
+
+  @action
+  onDataProviderChange = async (
+    dataProviderId = this.rootStore.dataProvider
+  ) => {
+    const statistics = await this.retrieveStatistics(dataProviderId)
+    Object.assign(this, statistics)
+  }
+}
+
+export default Statistics
--- a/pages/data-providers/[data-provider-id]/overview.jsx
+++ b/pages/data-providers/[data-provider-id]/overview.jsx
@@ -9,9 +9,14 @@ import { Button, Card, Overlay, Numeral } from 'design'
 // eslint-disable-next-line
 import styles from './overview.css'
 
+const formatNumber = (n, { locale = 'en-GB' } = {}) =>
+  n > 1000
+    ? `${(Math.floor(n / 10) / 100).toLocaleString(locale)}\xa0K`
+    : n.toLocaleString(locale)
+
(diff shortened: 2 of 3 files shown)


Joozty:
I would use `autorun` for it. Something similar as it is done [here](https://github.com/oacore/dashboard/blob/master/store/store.js#L46).


```
class Statistics {
  @observable statistic = {}

  constructor(rootStore) {
    this.rootStore = rootStore

    this.ononDataProviderChange()
  }
  
  retrieveStatistics = async () => {
    try {
      const [data] = await apiRequest(
        `/data-providers/${this.rootStore.dataProvider}/statistics`,
        'GET',
        {},
        true
      )
	  this.statistics = data 
    } catch (e) {
      if (e.statusCode !== 404) throw e
    }
  }
  
  onDataProviderChange = () =>
    autorun(() => {
		this.retrieveStatistics()
    })
}
```


Or you can use retrieveStatistics method in componentDidMount.

Joozty:
When you use autorun. Request gets fired even when you are not on that page. I want to fix it in the future. But for now, we need to live with it.

viktor-yakubiv:
> When you use autorun. Request gets fired even when you are not on that page. I want to fix it in the future. But for now, we need to live with it.

It's fine. With some endpoints we will need to do it anyway.

Thank you for your advice.

api/index.js
```diff
@@ -5,7 +5,7 @@ import axios from 'axios'
 import { NetworkError } from './errors'
 
 const CORE_API = 'https://api.core.ac.uk/internal'
-const CORE_API_DEV = 'https://api.dev.core.ac.uk/internal'
```
Joozty: Not a good idea. A cookie will not be sent.

store/statistics.js
```diff
@@ -0,0 +1,31 @@
...
+
+class Statistics {
+  constructor(rootStore) {
+    this.rootStore = rootStore
+  }
+
+  retrieveStatistics = async dataProviderId => {
+    try {
+      const { data } = await apiRequest(
+        `/data-providers/${dataProviderId}/statistics`,
```
Joozty: I was using it like this in my actions:

`/data-providers/${this.rootStore.dataProviderId}/statistics`

store/statistics.js
```diff
@@ -0,0 +1,31 @@
...
+      )
+
+      return data || {}
+    } catch (e) {
+      if (e.statusCode !== 404) throw e
+      return {}
+    }
+  }
+
+  onDataProviderChange = async dataProviderId => {
```
Joozty: would be nice to mark this as an `@action`

api/index.js
viktor-yakubiv: Yeah, I know 😔

store/statistics.js
viktor-yakubiv: Please, check the current solution. Does it satisfy you?

I try to make functions as pure as possible. Maybe here it's not the best what I could do. Let me know.

store/statistics.js
```diff
@@ -0,0 +1,43 @@
...
+      return {}
+    }
+  }
+
+  @action
+  onDataProviderChange = async (
+    dataProviderId = this.rootStore.dataProvider
+  ) => {
+    const statistics = await this.retrieveStatistics(dataProviderId)
+    Object.assign(this, statistics)
```
viktor-yakubiv: @Joozty I saw you used similar approach. Please, check in line 14 how I solved the issue with this. Line from the MobX's documentation:
> Only plain objects will be made observable. For non-plain objects it is considered the responsibility of the constructor to initialize the observable properties. Either use the `@observable` annotation or the `extendObservable` function.

Issue #75, opened 2020-01-16
Update deposit dates format


viktor-yakubiv:
On the deposit dates page we should display 2 fields:
- Publication date
- Public release date (aka Deposit date)

Both dates should be absolute. Additionally we can add a tooltip with relative format but it does not make any sense to user so it's not really necessary.

PR #90, opened 2020-01-23
Add publicDate column to deposit dates table

Joozty:
Closes https://github.com/oacore/dashboard/issues/75


pages/data-providers/[data-provider-id]/content.jsx
```diff
@@ -18,7 +18,6 @@ const tableConfig = {
     {
       id: 'oai',
       display: 'OAI',
-      sortable: true,
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/content.jsx]: This should not be part of commit that pretends to add something.

PR #92, opened 2020-01-24
Debugging login flow

viktor-yakubiv:
@Joozty I face the problem with URLs here and would be happy to hear your advice.

Store becomes first source of truth. It initialises from route first time and then all routes should reflect state of store.

Currently we have 2 redirect pages. `/` (root) index and `/data-providers` index. I think we have to remove these pages and handle routes somehow through store. The problem is that we have to handle other routes somehow through store what makes Next's implicit routing completely useless.

Do you have any opinion?


Joozty:
> Store becomes first source of truth. It initialises from route first time and then all routes should reflect state of store.

It works like that now there is no need to do any change, is there? Redirection happens on client-side. `componentDidMount` in _app.jsx is called only once on the client-side. We also [check whether root store has been initialized or not](https://github.com/oacore/dashboard/blob/master/store/index.jsx#L8).

store/user.js
```diff
@@ -1,81 +1,36 @@
...
-        {},
-        {},
-        true
-      )
-      this.repositories = data.organisation.repositories
-      this.selectedRepositoryId = this.repositories[0].id
-    } catch (e) {
-      if (e.status === 404) throw e
-    }
+    extendObservable(this, data)
```
Joozty: 👍

viktor-yakubiv:
@Joozty I managed to make this work. I will prettify commits in next hour or 2.

viktor-yakubiv:
@Joozty it's kind of final stage. Please, read commit messages.

store/root.js
```diff
@@ -0,0 +1,88 @@
...
+    )
+
+    if (dataProvider == null) {
+      throw new Error(
+        `User ${this.user.id} does not have access to DataProvider ${dataProviderId}.`
+      )
+    }
+
+    this.dataProvider = dataProvider
+    this.changeActivity('overview')
```
Joozty: IMHO user should stay on the page where they were.

viktor-yakubiv: No. If data provider changes users starts from home. Secondary navigation does not exist when primary navigates.

pages/_app/index.jsx
```diff
@@ -21,12 +26,40 @@ class App extends NextApp {
...
+
+    // Assumes store object is always the same
+    this.store = store
+    this.disposeRouter = autorun(this.reflectStoreToRoute.bind(this))
+
+    try {
+      const { dataProvider, activity } = new Route(window.location.pathname)
+      await store.init(dataProvider, activity)
+      this.setState({ isAuthorized: true })
+    } catch (unauthorizedError) {
```
Joozty: This error never will never happen I think.

pages/_app/index.jsx
```diff
@@ -21,12 +26,40 @@ class App extends NextApp {
...
+    // TODO: This probably should be under condition
+    router.push(route.href, route.as)
+  }
+
+  async componentDidMount() {
+    const store = initStore()
+
+    // Assumes store object is always the same
+    this.store = store
+    this.disposeRouter = autorun(this.reflectStoreToRoute.bind(this))
```
Joozty: I always considered using autorun outside of the store like a bad practice but I don't know why. :D

pages/_app/index.jsx
viktor-yakubiv: Yes but it should. If user is not authorised, cookie changes or something, we get 500. It's more API problem, we can expect 401.

store/deposit-dates.js
```diff
@@ -46,9 +42,9 @@ class DepositDates {
     return Math.round(level * 100) / 100
   }
 
-  retrieveDepositDates = async (pageNumber, searchTerm, columnOrder) => {
+  async retrieveDepositDates(pageNumber, searchTerm, columnOrder) {
     const order = getOrder(columnOrder)
-    const key = `${this.rootStore.dataProvider}-${pageNumber}-${searchTerm}-${order}`
+    const key = `${pageNumber}-${searchTerm}-${order}`
```
Joozty: I don't think this was a good move. If you switch dataProvider you will see wrong data.

pages/_app/index.jsx
viktor-yakubiv: In my opinion, `autorun` is a simple `observer`. Comparing to `computed` that runs on request, `autorun` runs always.

Maybe you have a point. I will think about it more.

store/deposit-dates.js
viktor-yakubiv: No. I recreate child stores when data provider changes.

pages/_app/index.jsx
Joozty: Sorry I didn't notice this return statement.

https://github.com/oacore/dashboard/blob/login-debugging/api/index.js#L34

viktor-yakubiv: Honestly, I forgot to remove that part and throw error instead.

viktor-yakubiv: I've just did it.

store/root.js
```diff
@@ -0,0 +1,88 @@
...
+  @observable works = null
+
+  @observable depositDates = null
+
+  @computed
+  get dataProviders() {
+    return this.user.dataProviders
+  }
+
+  @action async init(dataProviderId, activityPath) {
```
Joozty: I was thinking whether this should be marked as an action or not. Do you have any opinion about it?

[From docs](https://mobx.js.org/refguide/action.html):

> Actions should only, and always, be used on functions that modify state. Functions that just perform look-ups, filters etc should not be marked as actions; to allow MobX to track their invocations.

I haven't found a big advantage in using actions yet. I guess it's good for debugging store state with [Mobx Devtools](https://github.com/mobxjs/mobx-devtools)

viktor-yakubiv: I marked with `@action` each function that modifies the state according to the doc. If function calls the function that modifies the state, it's an action too. I am not sure about the second but agree to the first one.

PR #93, opened 2020-01-27
Logout

Joozty:


viktor-yakubiv:
Can we fallback to simple _Logout_ link without anything else? It might even better if it stays pure link to `API/logout`

Joozty:
@viktor-yakubiv please have a look now.

store/root.js
```diff
@@ -47,6 +47,13 @@ class Root {
     }
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  async logout() {
+    // TODO: /logout should ideally be moved to /internal
+    await apiRequest('https://api.dev.core.ac.uk/logout')
+    window.location.replace('/logout.html')
+  }
```
viktor-yakubiv: Do you add this method just because to do not like having link to API in the UI?

components/layout/app-bar.jsx
```diff
@@ -25,8 +26,14 @@ const DashboardAppBar = ({ children, className, ...restProps }) => {
       </AppBar.Brand>
 
       {children}
+      <AppBar.Item>
+        {`${store.user.givenName} ${store.user.familyName} `}|{' '}
+        <Button className={styles.logoutButton} onClick={store.logout}>
+          Logout
+        </Button>
+      </AppBar.Item>
```
viktor-yakubiv: Can you write this as a separate component that will be connected to store but the AppBar will not?

components/layout/app-bar.jsx
```diff
@@ -25,8 +26,14 @@ const DashboardAppBar = ({ children, className, ...restProps }) => {
       </AppBar.Brand>
 
       {children}
+      <AppBar.Item>
+        {`${store.user.givenName} ${store.user.familyName} `}|{' '}
```
viktor-yakubiv: Could you pleas add here some `span` with class that would hide on mobile screens?

components/layout/app-bar.jsx
```diff
@@ -25,8 +26,14 @@ const DashboardAppBar = ({ children, className, ...restProps }) => {
...
+        <Button className={styles.logoutButton} onClick={store.logout}>
```
viktor-yakubiv: I would prefer this being pure link.

Joozty:
@viktor-yakubiv That's why I created dropdown menu.

components/layout/app-bar.jsx
Joozty: How? It cannot be a pure link because the cookie won't be deleted.

store/root.js
Joozty: I don't mind having API in UI. But in that case, logout has to work like this `https://api.dev.core.ac.uk/logout?continue=someurl`. 

Also, it's nice to have some action in the store. For example, when we store something in local storage this is the right place for removing it from there.

components/layout/app-bar.jsx
viktor-yakubiv: Link to `api.core.ac.uk/logout?continue=/logout.html`

components/layout/app-bar.jsx
Joozty: sure

store/root.js
viktor-yakubiv: I agree. On the other hand, having another action that can be avoided is not that good. Less code is less errors what it better.

Joozty:
@viktor-yakubiv I did changes you requested. As you said we should distinguish API URL based on NODE_ENV. I leave it as is for now. It should be handled in a separate PR. 

NOTE: It doesn't work now. I think @valerapro will deploy that continue param tomorrow.

components/application/logout.jsx
```diff
@@ -7,7 +7,7 @@ import { withGlobalStore } from 'store'
 
 const LOGOUT_URL = `https://api.dev.core.ac.uk/logout?continue=${encodeURI(
   typeof window !== 'undefined'
-    ? `${window.location.protocol}//${window.location.host}`
+    ? `${window.location.protocol}//${window.location.host}/logout.html`
```
viktor-yakubiv: You may find useful [`window.location.origin`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/origin)

components/application/logout.jsx
```diff
@@ -7,7 +7,7 @@ import { withGlobalStore } from 'store'
 
 const LOGOUT_URL = `https://api.dev.core.ac.uk/logout?continue=${encodeURI(
```
viktor-yakubiv: Perhaps, you meant [`encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)

components/application/logout.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+const LOGOUT_URL = `https://api.dev.core.ac.uk/logout?continue=${encodeURI(
+  typeof window !== 'undefined'
+    ? `${window.location.protocol}//${window.location.host}/logout.html`
+    : ''
+)}`
+
+const Logout = ({ store }) => (
+  <AppBar.Item className={styles.container}>
+    {`${store.user.givenName} ${store.user.familyName} `}|{' '}
```
viktor-yakubiv: Can this be `<span className={styles.name}>{name}</span>`?

where 

```js
name = `${store.user.givenName} ${store.user.familyName}` || store.user.email`
```

and

```css
.name {
  &::after { content: ' | ' }
  @media (--screen-until-desktop) { display: none }
}
```

PR #94, opened 2020-01-27
Dropdown menu

Joozty:


viktor-yakubiv:
Let's postpone this please.

Joozty:
@viktor-yakubiv why?

viktor-yakubiv:
Because it does not look perfectly and causes bugs. I am not sure in the interface and all of this may block development. We should more focus on valuable but not awesome product so it may be better to fall back to pure link for logout and do this slightly later.

PR #95, opened 2020-01-28
Connect Deposit Lag chart to API

Joozty:


pages/data-providers/[data-provider-id]/deposit-dates.jsx
```diff
@@ -119,7 +119,12 @@ const DepositDates = ({
         <p>
           The chart displays work distribution per deposit time lag in days.
         </p>
-        <TimeLagChart data={store.depositDates.timeLagData} />
+        {Boolean(store.depositDates.timeLagData.length) && (
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-dates.jsx]: It's better to write `length > 0`. Check Airbnb's JavaScript Style Guide [22.6](https://github.com/airbnb/javascript#coercion--booleans) and [15.3](https://github.com/airbnb/javascript#comparison--shortcuts)

pages/data-providers/[data-provider-id]/deposit-dates.jsx
```diff
@@ -119,7 +119,12 @@ const DepositDates = ({
...
+          <TimeLagChart data={store.depositDates.timeLagData} />
+        )}
+        {!store.depositDates.timeLagData.length &&
+          !store.depositDates.isRetrieveDepositDatesInProgress &&
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-dates.jsx]: I would simplify it to either `depositDates.loading` or `depositDates.requests > 0` (`depositDates.requestsInProgress > 0`).

To be fair, the data can be in loading, present or absent, so there are 2 levels of conditions:
- if loading, show loading
- else if data show data
- else show message

pages/data-providers/[data-provider-id]/overview.jsx
```diff
@@ -87,6 +87,9 @@ const DataStatisticsCard = ({
 
 const DepositingCard = ({ chartData, complianceLevel }) => {
   const { title, description, action } = depositing
+
+  if (chartData.length === 0) return null
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/overview.jsx]: Should not be here. It's not a problem of card. It's more about the page.

store/deposit-dates.js
```diff
@@ -19,7 +19,7 @@ class DepositDates {
 
   @observable isRetrieveDepositDatesInProgress = false
 
-  timeLagData = []
+  @observable timeLagData = []
```
viktor-yakubiv: Did you mean _too_? 🙂

pages/data-providers/[data-provider-id]/deposit-dates.jsx
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/deposit-dates.jsx]: These 2 are mutually excluded but to improve readability I would prefer `?:` at least operator.

However, ideally it would be `isLoading ? 'It\'s loading' : (data.length > 0) ? <Chart /> : 'No data')`

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-dates.jsx]: Eslint complains about nested ternary operators. That's why I did two separate statements.

store/deposit-dates.js
Joozty: yes

pages/data-providers/[data-provider-id]/overview.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/overview.jsx]: What do you propose? Do you rather want to show a card without chart? If so it will show wrong data in the description. Should I replace `100%` -> `unknown.`

```
You have about 100% non-compliant works.
```

pages/data-providers/[data-provider-id]/deposit-dates.jsx
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/deposit-dates.jsx]: You can write it all separate but please add loading state reposting if you already started in the beginning.

pages/data-providers/[data-provider-id]/overview.jsx
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/overview.jsx]: You can do it one level above, in `DashboardView` component. 

Either create placeholder card that would display _We are collecting your data. It will appear soon here_ and display it if data absent or add such test into the card and show text inside. 

Despite repeating some parts I would go first way.

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/overview.jsx]: https://github.com/oacore/dashboard/pull/95/commits/c9b79e5237a71db01f58645a6a9f349a74b0bf38

texts/depositing/overview.yml
```diff
@@ -1,3 +1,14 @@
...
+     We are collecting data. It should appear anytime soon.
+  {{/isLoading}}
+  {{^isLoading}}
+    {{#count}}
+       You have about {{amount}} non-compliant works.
+    {{/count}}
+    {{^count}}
+       We couldn't find any data about depositing performance.
+   {{/count}}
+  {{/isLoading}}
```
viktor-yakubiv: `@oacore/texts` was designed for a content manager to be able edit texts with minimal involvement of developer. I do not think that somebody not-technical would learn another technology just because of this. Such text is not clear enough to easily edit by other people.

To be fair, I understand it only because I know the task. I am not familiar with Mustache enough to know that it's conditional rendering.

Joozty:
@viktor-yakubiv  Go ahead if you think you can make it better...

viktor-yakubiv:
@Joozty could you please check [this one](https://github.com/oacore/dashboard/pull/95/commits/ba1f724d10e5d87329cc2b527b6e50106130ee13)? I am not sure about using `null` to check data existence. It seems to be clear but not clear enough. What do you think?

Joozty:
it's fine. feel free to merge

PR #98, opened 2020-01-28
Fix GET params & HEAD method in API request

viktor-yakubiv:
1. Improves URL checking
2. Fixes GET parameter passing
3. Ignores response body for HEAD methods

From #97


viktor-yakubiv:
I wanted you to add `searchParams` improvements here but anyway. Thank you

Issue #96, opened 2020-01-28
Disable Take down buttons


viktor-yakubiv:
We should taking down buttons for the first time. Buttons do not report any actual behaviours, do not send requests.

PR #100, opened 2020-01-29
Disable Take down buttons

Joozty:
Closes https://github.com/oacore/dashboard/issues/96

### DIFF
--- a/components/published-toggle/index.css
+++ b/components/published-toggle/index.css
@@ -5,11 +5,13 @@
   --toggle-active-color: #5a9216;
   --toggle-switch-border-radius: 2px;
 }
+
 .toggle {
   position: relative;
   display: inline-block;
   cursor: pointer;
 }
+
 .toggle-checkbox {
   position: absolute;
   visibility: hidden;
@@ -85,4 +87,9 @@
 .toggle-switch::after {
   color: var(--gray-100);
   mask-image: var(--icon-unpublished);
+}
+
+.disabled {
+  cursor: not-allowed;
+  opacity: 0.5;
 }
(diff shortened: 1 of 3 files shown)


Joozty:
Ohh sorry for the misunderstanding. It should be fine now.

components/published-toggle/index.css
```diff
@@ -5,11 +5,12 @@
   --toggle-active-color: #5a9216;
   --toggle-switch-border-radius: 2px;
 }
+
 .toggle {
   position: relative;
   display: inline-block;
-  cursor: pointer;
```
viktor-yakubiv: Let's keep it.

components/published-toggle/index.jsx
```diff
@@ -1,13 +1,12 @@
-import React, { useState } from 'react'
+import React from 'react'
 import classNames from 'classnames'
 
 import styles from './index.css'
 
 const PublishedToggle = ({ className, defaultVisibility, isExpanded }) => {
```
viktor-yakubiv: Could you please provide `disabled` prop here and just disable it from parent?

Disabled toggle should have transparent background and different cursor.

```css
.disabled {
  opacity: 0.5; /* or any other */
  cursor: not-allowed;
}
```

components/published-toggle/index.jsx
```diff
@@ -14,22 +19,29 @@ const PublishedToggle = ({ className, defaultVisibility, isExpanded }) => {
...
-          [styles.expanded]: isExpanded,
-        })}
+        className={classNames
+          .use('toggle-checkbox', {
+            expanded: isExpanded,
+          })
+          .from(styles)}
         type="checkbox"
         onChange={event => {
+          if (disabled) return
```
viktor-yakubiv:
```suggestion
          if (disabled) {
            event.preventDefault()
            return
          }
```

PR #102, opened 2020-01-30
Sentry integration

Joozty:
Example of error with user metadata [here](https://sentry.io/organizations/oacore/issues/1484745460/?project=2091955&query=is%3Aunresolved)

Last commit has to be dropped before merge. It's just demenstration of error boundary message.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+SENTRY_DSN=
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -14,3 +14,6 @@ settings:
     node:
       paths: ['.']
       extensions: ['.js', '.jsx']
+
+globals:
+  SENTRY_DSN: true
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ build
 node_modules
 public/design
 public/icons.svg
+.env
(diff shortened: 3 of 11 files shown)


viktor-yakubiv:
Does it work?

Joozty:
Yes, why? (Sentry is disabled for now.sh and localhost)

viktor-yakubiv:
Just clarifying.

I would like to avoid `dotenv` if favour of present/absent environmental variable. Everything else looks good to me.

Could you please remove the last testing commit and then feel free merge? Or do you want to push that commit too?

Joozty:
> I would like to avoid dotenv if favour of present/absent environmental variable. Everything else looks good to me.

I wouldn't do that. When I want to test something on localhost it's easier for me to change .env file instead of exporting my env variable. I may even forget to "unexport" that variable after a test. Besides, you don't need to use that. You can freely export your variables. Also dotenv doesn't affect page size...

viktor-yakubiv:
It's matter of taste. I prefer one, you another. It's okay.

However, your `.env` file is ignored so you can only know it exists or has been changed if you manually check it. It has the same cons as simple exporting. 

By the way, you export for a single terminal session. Because it does not persist you have fresh environment for each time you close and open terminal window.

Joozty:
> By the way, you export for a single terminal session. Because it does not persist you have fresh environment for each time you close and open terminal window.

Unfortunately, I don't run it in terminal. I have it configurated in my IDE and if I need to change it I need to go to configuration settings and change ENV variables there. This is much simpler for me but I agree that .env is not tracked and I may forget to remove it from there.

PR #104, opened 2020-01-30
Show warning message when wrong credentials passed

Joozty:


Joozty:
@viktor-yakubiv Not sure if we want to have some close button there or make message disappear after some time.

public/login.html
```diff
@@ -80,3 +81,18 @@ <h1>Login</h1>
...
+<script>
+  function login(e) {
+    e.target.action = 'https://api.dev.core.ac.uk/login_check?continue=' + encodeURIComponent(
+      window.location.origin
+    );
+  }
+
+  const urlParams = new URLSearchParams(window.location.search);
+  if (urlParams.has('reason')) {
+    const message = document.getElementById('error-message')
```
viktor-yakubiv: I am considering using more old JS but whatever works. I am not sure about our browsers list yet. We actually support only latest Chrome but we will need to wide this up.

Maybe `var` and avoiding `URLSearchParams` in favour of pure regular expression test would work better.

public/login.html
```diff
@@ -80,3 +81,18 @@ <h1>Login</h1>
...
+    message.innerHTML='The username or password you entered is incorrect.';
+    message.style.display = 'block';
```
viktor-yakubiv: Simple CSS snippet may help you avoid this.

```css
.message:empty { display: none }
```

Joozty: This is really cool selector. Thanks

PR #107, opened 2020-01-31
Use correct publication date data field

viktor-yakubiv:
I display primarily repository's publication date and if it's empty, Crossref's one. @mcancellieri let me know if I have to change this logic, you know the data better.

I would be happy to display both fields but for this moment I may affect layout negatively. Let's do it when we fix the layout please.

---

@Joozty we merge the PR only when @mcancellieri approves changes.


viktor-yakubiv:
Closing in favour of calculating `publicationDate` on the API side. The algorithm is too complex so it's better to calculate in a single place and expose.

PR #108, opened 2020-01-31
Profiles

Joozty:
@viktor-yakubiv Let me explain the idea behind this. I was thinking about distinguishing dev and production API URLs. Obviously we could do it simply by checking env variable. But I decided to propose something different - changing API dynamically.

1. Three profiles are created localhost, staging, and production and you can switch between them dynamically.
2. The localhost profile may be quite powerful especially for our backend guys. They don't need to push bugfixes and deploy to dev to check whether it was fixed or not.
3. The profile is stored in session storage which means it gets removed after tab is closed. No need to switch back to the default profile. 
4. No need to restart the server to use different API URL.
5. We can use profiles for storing text/images CDN too.

Please let me know what do you think. This is more like a proposal. It doesn't have to be merged. Also, it's not finished. I need to find a better place for profile switch also it should be shown only for admin users. I just need to know if it is worth to continue.

 

NOTE: You will need to run this code sample on localhost for the time being. now.sh is considered as a production env. I need to change it.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -17,3 +17,4 @@ settings:
 
 globals:
   SENTRY_DSN: true
+  PROFILE_DATA: true
--- a/api/index.js
+++ b/api/index.js
@@ -1,26 +1,19 @@
 import { NetworkError } from './errors'
 
-const CORE_API = 'https://api.core.ac.uk/internal'
-const CORE_API_DEV = 'https://api.dev.core.ac.uk/internal'
-
 const apiRequest = (
   url,
   method = 'GET',
   queryParams = {},
-  customHeaders = {},
-  dev = false
+  customHeaders = {}
 ) => {
   const controller = new AbortController()
   const { signal } = controller
 
-  const requestUrl = new URL(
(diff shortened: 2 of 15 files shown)


viktor-yakubiv:
It might help @valerapro and @mcancellieri test API without running front-end themselves. Even if this is incorrect form the design perspective we can temporarily find this useful.

Could you please find a way to include profiles to build only when `NODE_ENV=development` or any other environmental variable exists if you are talking only about API address.

---

Also, please consider one more simple solution: passing environment variable like `CORE_API_ROOT_URL` that can be whatever or production API URL if missing and then:

- if you want to use different API locally, just set the variable
- build `dashboard.dev.core.ac.uk` with dev version of API
- build *.oacore.now.sh with production version of API
- build production with production (there is no other way)

The only problem I see, we have to populate this variable to static HTML pages which should not be built by any tool.

viktor-yakubiv:
Is not this outdated?

Joozty:
It is. I am gonna close it. As only I was excited about this feature. 😄

PR #109, opened 2020-01-31
Repository select search

Joozty:


store/user.js
```diff
@@ -9,6 +9,12 @@ class User {
     return this.dataProviders.some(({ id }) => dataProviderId === id)
   }
 
+  searchDataProviders(searchTerm) {
+    return this.dataProviders.filter(
+      e => e.name.toLowerCase().search(searchTerm.toLowerCase()) !== -1
```
viktor-yakubiv: Are you sure it's not `.indexOf`? According to the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search), `.search()` requires regular expression.

I would use `>= 0` to compare but it's up to you.

Joozty: > If a non-RegExp object regexp is passed, it is implicitly converted to a RegExp with new RegExp(regexp).

viktor-yakubiv: It's not a deal but I think `indexOf` would work faster.

Joozty: That's true. I just forgot it does exist.

PR #122, opened 2020-02-05
Extend overview chart data to show 4 years period, make it relative to data

viktor-yakubiv:
Deposit time lag chart on Overview page moves compliance limit point (90 days) relatively to compliance level percentage.

For example, for a compliance level 80% first 80% of chart is green and last 20% are grey.

Also, the limit of filtered data for overview extended to 4 years.


viktor-yakubiv:
@PMRuma could you please test this feature?

PR #126, opened 2020-02-05
Force using production API if NODE_ENV is production

viktor-yakubiv:
I needed to add this quick dirty fix up to API client to make it work project-wide.

The fix makes `apiRequest` function ignore `dev` parameter. @Joozty is working on better version of checking environments in #108.

Passing correct URL to login page is still a question. I had to replace 2 strings but we do not support login to dev API anymore. I am considering having 2 equal pages for different environments but I will definitely think about it more.

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -1,7 +1,15 @@
 import { NetworkError } from './errors'
 
 const CORE_API = 'https://api.core.ac.uk/internal'
-const CORE_API_DEV = 'https://api.dev.core.ac.uk/internal'
+
+/**
+ * Forces using production API URL and ignore `dev` parameter
+ * in the function below.
+ */
+const CORE_API_DEV =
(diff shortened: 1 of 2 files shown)


viktor-yakubiv:
@Joozty did you test this? I could not log in yesterday

Joozty:
I didn't test it. Seems like we didn't reflect all changes from dev -> prod


```
Access to fetch at 'https://api.core.ac.uk/internal/user' from origin 'https://dashboard-git-remove-dev-api.oacore.now.sh' has been blocked by CORS policy: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.
```

viktor-yakubiv:
@mcancellieri could you please take a look at it if you have time?

PR #129, opened 2020-02-05
Revert bump core-js from 3.4.2 to 3.6.4

viktor-yakubiv:
Reverts oacore/dashboard#110


Joozty:
Let's wait for node upgrade and if it does not help you can merge this.

viktor-yakubiv:
Node upgrade does not work. We should downgrade `core-js`

PR #136, opened 2020-02-07
Remove ordering by publicReleaseDate

viktor-yakubiv:
Quick fix up.


viktor-yakubiv:
It does not have any affect. I am closing it

PR #140, opened 2020-02-13
CSP

Joozty:
You may find more info in commits. I am not sure about the last one. It makes our app more secure but on the other hand it is very strict.


pages/_app/index.jsx
```diff
@@ -36,6 +36,29 @@ class App extends NextApp {
...
+        'Content-Security-Policy',
+        [
+          // consider everything from these two domains as a safe
+          "default-src 'self' *.core.ac.uk",
+          // in development there are attached inline scripts
+          // (probably from hot reload or some Next.JS magic)
+          `script-src 'self' *.google-analytics.com ${
+            process.env.NODE_ENV !== 'production' ? "'unsafe-inline'" : ''
+          }`,
+          // TODO: It's not a big deal but we should ge rid of
```
viktor-yakubiv:
```suggestion
          // TODO: It's not a big deal but we should get rid of
```

pages/_app/index.jsx
```diff
@@ -36,6 +36,29 @@ class App extends NextApp {
...
+          //  all inline styles if possible as it is not recommended
```
viktor-yakubiv: It's not really possible. We will need to work around markdown custom styling somehow.

public/login.css
```diff
@@ -0,0 +1,58 @@
+html,
+body {
+  height: 100%;
+}
+body {
```
viktor-yakubiv: Should contain an empty line. Does not linter check this file?

public/login.js
```diff
@@ -0,0 +1,14 @@
+function login(e) {
+  e.target.action = `https://api.core.ac.uk/login_check?continue=${encodeURIComponent(
+    window.location.origin
+  )}`
+}
+
+const urlParams = new URLSearchParams(window.location.search)
```
viktor-yakubiv: ~Do you ming using `var`?~ Sorry, you are using new Web API, that we do not polyfill. Never mind.

Also, could you please what the script into DOM loaded event listener?

public/login.css
viktor-yakubiv: If you add this to linter, consider moving styles into subdirectory where you lint everything.

public/login.js
Joozty: > Also, could you please what the script into DOM loaded event listener?

I don't get it sorry.

viktor-yakubiv: To have save save JS execution you should make DOM operations e.g. `Element.addEventListener` after having actual DOM parsed, i.e. after on load event. Check [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event) event to find more.

public/login.css
Joozty: It should be processed. Let me check it. 

Btw. I ran into a problem where stylelint doesn't respect ignore comments, i.e. if you use `--fix` option than `sanitize.css` file gets processed and auto fixed. 😞 Maybe sanitize.css should be minimized with webpack or we should have it unminified.

pages/_app/index.jsx
viktor-yakubiv: It's used fairly rarely. I guess once at this moment but it will happen.

public/login.css
viktor-yakubiv: It's fine to have `sanitize.js` fixed by linter. We do not have problems with network load there.

By the way, did you changed to local version because of content policy you wrote? Did not you want to add CDN and continue using Sanitize from there?

public/login.js
```diff
@@ -0,0 +1,16 @@
+function login(e) {
+  e.target.action = `https://api.core.ac.uk/login_check?continue=${encodeURIComponent(
```
viktor-yakubiv: If you are extracting JS out of the HTML file, could you please rely on `action` attribute as single source of truth to avoid repeating and potential issue in the future if we decide to change the URL?

viktor-yakubiv: You can use `URL` for that. You have `URLSearchParams`, so we expect to have URL Web API anyway.

public/login.css
Joozty: > By the way, did you changed to local version because of content policy you wrote? Did not you want to add CDN and continue using Sanitize from there?


I changed it because of security. It may happen that sanitize.css would contain malicious code. I know it's more than impossible but still, we should be cautious especially when dealing with credentials 😄. If I didn't read the article you sent me I would probably leave it there as it was.

Joozty: https://github.com/oacore/dashboard/pull/140/commits/85d06cb1bd05da4d05942e1030e2975910f8593e

PR #141, opened 2020-02-18
Allow to copy text in table

Joozty:
@viktor-yakubiv I hope this is sufficient or did you imagine a more sophisticated solution, i.e. auto copy to the clipboard?

### DIFF
--- a/components/table/index.jsx
+++ b/components/table/index.jsx
@@ -112,6 +112,10 @@ class InfiniteTable extends React.PureComponent {
     const { expandable } = this.props
     if (!expandable) return
 
+    // ignore copy event
+    const selection = window.getSelection().toString()
+    if (selection.length) return
+
     const clickedRow = event.target.closest('tr')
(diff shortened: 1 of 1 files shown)


components/table/index.jsx
```diff
@@ -109,23 +111,36 @@ class InfiniteTable extends React.PureComponent {
   }
 
   handleRowClick = event => {
-    const { expandable } = this.props
-    if (!expandable) return
+    if (event.detail === 1) {
```
viktor-yakubiv: I do not see any way this `1` comes here and what it means. Detail technically should not but meaningfully should be an object.

Joozty: This is a number of times element is clicked. See [docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event). It does work very well but it's good to have it there.

viktor-yakubiv: Thank you. I did not know that 🙂

PR #142, opened 2020-02-21
Fix development user up

viktor-yakubiv:
@Joozty please consider this quick fix up as development improvement. Although, please take a closer look because it does not seem to work.


viktor-yakubiv:
I see no benefit in that too. Thank you @Joozty

PR #144, opened 2020-02-21
Disable Content-Security-Policy in development environment

viktor-yakubiv:
Images are not shown because of Content-Security-Policy, at least on localhost. I propose to disable it in development environment.

I may be wrong but I cannot display an image from CORE static website neither on localhost nor on [staging](https://dashboard-git-installation-instructions.oacore.now.sh/data-providers/27/plugins/discovery).

It affects #87


viktor-yakubiv:
Thank you @Joozty I could not even notice that.

PR #146, opened 2020-02-22
Move data fetch from Table to Store

Joozty:
@viktor-yakubiv This is a prerequisite for takedown article feature in the content table. From now on table component keeps only offset and size values.  Data are stored in the store. This will allow changing the `visibility` value of a particular row.

I still don't know how to handle page reload. Currently, the API endpoint is cached which means after hard reload table component will show wrong data. We can bypass the cache via fetch `{cache: "reload"}` but this will bring a new issue. It won't load the same rows because of elasticsearch. :/


store/helpers/pages/pages.js
```diff
@@ -0,0 +1,97 @@
...
+import Page from './page'
+import invalidatePreviousRequests from '../invalidatePreviousRequests'
+
+import { LinkedList } from 'utils/linkedList'
+import apiRequest from 'api'
+import { CancelablePromise } from 'utils/promise'
+
+const PAGE_SIZE = 100
+
+export default class Pages {
```
viktor-yakubiv: Try always export explicitly in the end of file in a single place. Sometimes you might want to add a decorator to the class and this will affect more lines than it actually should. Also, export in a single point improve readability.

store/helpers/pages/pages.js
```diff
@@ -0,0 +1,97 @@
...
+  #pages = new LinkedList()
+
+  #searchTerm = ''
+
+  #columnOrder = ''
+
+  #pageNumber = 0
```
viktor-yakubiv: I see no reason to use private properties here but as you wish.

store/helpers/pages/pages.js
```diff
@@ -0,0 +1,97 @@
...
+
+  isLastPageLoaded = false
+
+  isFirstPageLoaded = false
```
viktor-yakubiv: If you go to private properties, this technically should be a getter of another private property.

Sometimes JavaScript gives you and everybody a freedom that if you want to take off you should spend a lot of effort. I prefer to do not do it at all.

store/helpers/pages/pages.js
```diff
@@ -0,0 +1,97 @@
...
+
+  constructor(url) {
+    this.url = url
+  }
+
+  async slice(from, to) {
+    while (this.#pages.length < to && !this.isLastPageLoaded)
+      // eslint-disable-next-line no-await-in-loop
+      await this.load().promise
```
viktor-yakubiv: Interesting. I am afraid to do so. Be careful. Seems to cause unexpected behaviour

Joozty:
> I do not agree with what you say. It seems you tightly couple the store and the table what is not correct but the code does not look bad

That's partially true. It's not tight up to store but it's tight up to Pages class, i.e. table component will work only when you pass down [Pages](https://github.com/oacore/dashboard/blob/move-data-fetch/store/helpers/pages/pages.js#L11) instance. The truth is that Pages class is used only in the store and it's customized for its needs.


> Please, go ahead if you are confident in what you do. 
Well, I don't like it but it's the only way I can see now.

store/helpers/pages/pages.js
Joozty: It's fine to use it in a while loop. It wouldn't work if it is used in forEach callback.

store/helpers/pages/pages.js
Joozty: > If you go to private properties, this technically should be a getter of another private property.
I am not an expert in object-oriented design but I don't see any reason for doing so.

PR #148, opened 2020-02-25
Takedown

Joozty:


store/works.js
```diff
@@ -1,9 +1,31 @@
...
     super(url)
+    this.worksUrl = url
+  }
+
+  @action
+  async changeVisibility(rowId, isVisible) {
+    const [pageNumber, workId] = rowId.split('-', 2)
+    await apiRequest(
+      `${this.worksUrl}/${workId}`,
+      'PUT',
```
viktor-yakubiv: It should be PATCH because the update is partial. You can find the [classification on Wikipedia](https://en.wikipedia.org/wiki/Representational_state_transfer#Relationship_between_URI_and_HTTP_methods).

Joozty: Well, I used what was implemented. So if they change it on the backend I don't mind it to change here.

viktor-yakubiv: @valerapro is aware of it

Joozty:
Waiting for @valerapro to fix PATCH endpoint. It can be reviewed in the meantime.

api/index.js
```diff
@@ -24,6 +25,7 @@ const apiRequest = (
     method,
     credentials: 'include',
     headers: requestHeaders,
+    body: data && JSON.stringify(data),
```
viktor-yakubiv: If you add this feature, I recommend you to check the data type.

If `data` is an Object, we should stringify it to JSON, if text/number, only inline it. Also, based on data type we have to preset `Content-Type` header to either `application/json` or `text/plain` unless user defines it themselves.

components/published-toggle/index.jsx
```diff
@@ -1,10 +1,11 @@
-import React, { useState } from 'react'
+import React, { useState, useCallback } from 'react'
 import { classNames } from '@oacore/design/lib/utils'
 
 import styles from './index.css'
 
 const PublishedToggle = ({
   className,
+  onVisibilityChanged,
```
viktor-yakubiv: It's a form element. Simple `onChange` and `checked` would be enough.

components/published-toggle/index.jsx
```diff
@@ -16,6 +17,20 @@ const PublishedToggle = ({
     else toggletext = 'UnPublished'
   }
 
+  const handleChange = useCallback(async event => {
+    if (disabled) {
```
viktor-yakubiv: I see no reason for this. Change is never fired on disabled inputs, is it?

components/published-toggle/index.jsx
```diff
@@ -16,6 +17,20 @@ const PublishedToggle = ({
...
+      event.preventDefault()
+      return
+    }
+    const isVisible = event.target.checked
+    try {
+      await onVisibilityChanged(isVisible)
+      toggleChecked(isVisible)
+    } catch (e) {
```
viktor-yakubiv: Try to use self-explained variables.

```suggestion
    } catch (ignoredNetworkError) {
```

components/published-toggle/index.jsx
Joozty: It's designed like that for now. Internally it's just input element but outside it shouldn't look like that.

Joozty:
Just tested it and it's ok for me. @viktor-yakubiv wdyt?

components/published-toggle/index.jsx
```diff
@@ -16,6 +17,16 @@ const PublishedToggle = ({
     else toggletext = 'UnPublished'
   }
 
+  const handleChange = useCallback(async event => {
```
viktor-yakubiv: Why do you use `useCallback` hook here?

store/works.js
```diff
@@ -1,10 +1,29 @@
...
   constructor(baseUrl) {
     const url = `${baseUrl}/works`
     super(url)
   }
+
+  @action
+  async changeVisibility(rowId, isVisible) {
+    const [pageNumber, workId] = rowId.split('-', 2)
+    await apiRequest(
+      `/works/${workId}`, // TODO: consider moving this endpoint to /data-providers/works/
```
viktor-yakubiv: The URL should be discovered from Work entity by accessing an `.url` key.

pages/data-providers/[data-provider-id]/content.jsx
```diff
@@ -61,12 +61,14 @@ const tableConfig = {
...
       className: styles.visibilityColumn,
       getter: v => !v.disabled,
-      render: (v, isExpanded) => (
+      render: (v, { id, isExpanded }) => (
         <PublishedToggle
           className={styles.visibilitySwitch}
           defaultVisibility={v}
           isExpanded={isExpanded}
-          disabled
+          onVisibilityChanged={isVisible =>
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/content.jsx]: I think that if you have a wrapper around checkbox it should completely implement checkbox interface at least because checkbox interface is a standard.

components/published-toggle/index.jsx
```diff
@@ -16,6 +17,16 @@ const PublishedToggle = ({
...
+    const isVisible = event.target.checked
+    try {
+      await onVisibilityChanged(isVisible)
```
viktor-yakubiv: I think you should not do anything here, just call passed method.

Keep components being pure where it's possible.

components/published-toggle/index.jsx
Joozty: It's just a very small optimization so the function doesn't get recomputed on every rerender.

components/published-toggle/index.jsx
Joozty: it has to be done like this. Otherwise, if a request fails user may see the wrong data in table.

Joozty: Anyway we will have better interface once we implement your design. So I would pay much attention to this...

store/works.js
viktor-yakubiv: There is no `work.url` yet. My comment was only to reply your comment.

components/published-toggle/index.jsx
viktor-yakubiv: The new design may no happen anytime soon just because we do not have enough backend support for that. Also it requires finishing #94.

Is this comment affected by #179?

Joozty: >  we do not have enough backend support for that.

How so? We just need to rewrite how expanded row is shown.

viktor-yakubiv: If you check the [fresh design](https://www.figma.com/proto/rWs351bJa8uVOSS97Nk8Tm/Content?node-id=78%3A41957&viewport=-256%2C347%2C0.9490244388580322&scaling=min-zoom), there are a few bits that require backend support:
- having document issues right in the document entity because it's show contextually in the document metadata and right in the table
- having DOI source right in the document metadata

Merging everything into a single index takes a lot of effort on the backend.

Joozty: So I rebased the master. Can we proceed with the merge? I will keep is as is As I already said it's [not that easy to implement](https://github.com/oacore/dashboard/pull/179#discussion_r391007921).

PR #150, opened 2020-02-26
Fix CSP for styles

Joozty:
We need to use unsafe inline because Nextjs use inline styles for dev build watcher (icon at the right bottom of a page)

### DIFF
--- a/pages/_app/index.jsx
+++ b/pages/_app/index.jsx
@@ -13,6 +13,8 @@ import { initStore, GlobalProvider } from 'store'
 import Application from 'components/application'
 import { Sentry } from 'utils/sentry'
 
+import './index.css'
+
 process.on('unhandledRejection', err => {
   Sentry.captureException(err)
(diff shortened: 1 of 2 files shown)


pages/_app/index.jsx
```diff
@@ -13,6 +13,8 @@ import { initStore, GlobalProvider } from 'store'
 import Application from 'components/application'
 import { Sentry } from 'utils/sentry'
 
+import './index.css'
```
viktor-yakubiv: This may create a conflict with design, files order may be different and unpredictable. I would not export this to a file here, we already had some issues files order before.

Joozty: You are right but where else should I put it. I was thinking about `public/reset-styles.css` and load it manually. Do you think it would be better than this?

Joozty: Ok let's keep it like that for now.

PR #152, opened 2020-02-27
Allow to pass body to apiRequest

viktor-yakubiv:
Slightly improves code structure.

Adds description comments.

Comes from #148

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -7,25 +7,39 @@ const apiRequest = (
   url,
   method = 'GET',
   queryParams = {},
-  customHeaders = {}
+  customHeaders = {},
+  data
 ) => {
-  const controller = new AbortController()
-  const { signal } = controller
-
+  // Initialise URL
   const requestUrl = new URL(/^\w+:\/\//.test(url) ? url : `${API_URL}${url}`)
-  const requestHeaders = {
-    Accept: 'application/json',
-    ...customHeaders,
-  }
-
   requestUrl.search = new URLSearchParams(queryParams).toString()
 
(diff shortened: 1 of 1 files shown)


api/index.js
```diff
@@ -7,25 +7,39 @@ const apiRequest = (
...
+      Accept: 'application/json',
+    },
+  }
+
+  // Process body
+  if (method.toUpperCase() !== 'GET') {
+    if (typeof data == 'object') {
+      requestOptions.body = JSON.stringify(data)
+      requestOptions.headers['Content-Type'] = 'application/json'
+    } else requestOptions.body = data
```
Joozty: Don't you want to pass `text/plain` for this case?

api/index.js
```diff
@@ -7,25 +7,39 @@ const apiRequest = (
...
-    signal,
-  })
+    headers: {
+      Accept: 'application/json',
+    },
+  }
+
+  // Process body
+  if (method.toUpperCase() !== 'GET') {
+    if (typeof data == 'object') {
```
Joozty:
```suggestion
    if (typeof data === 'object') {
```

viktor-yakubiv: We use `eqeqeq: smart` for a reason. `typeof` always returns a string, so if you compare it to another string, there is no need to write another character.

api/index.js
viktor-yakubiv: I leave the responsibility to set correct content type up to the user.

However, I've just checked, `text/plain` is default and there is no need in setting it explicitly.

PR #153, opened 2020-02-28
Bump design

Joozty:


viktor-yakubiv:
I find it easier to [store icons in the project](https://github.com/oacore/dashboard/pull/87/files#diff-d32c48c7baa55db4d5494bd4ece42a3a) for this moment. Also, our [sidebar navigation](https://github.com/oacore/dashboard/blob/master/components/application/activity-select/index.jsx#L10) uses icons from the project because icons in the `design` had grey fill what made them not accept current font colour. I found difficult to do the release cycle because of this change.

To accept a6c3393 we have to fix such behaviour. However, I would keep the script an icons.

Joozty:
Isn't the time to moving some components and some logic to @oacore/design?

viktor-yakubiv:
Yes, it is but the move should not break anything

components/application/activity-select/index.jsx
```diff
@@ -7,7 +7,7 @@ import { navigation } from 'texts'
 
 const toUrl = value => (value === 'index' ? '.' : `./${value}`)
 const toIcon = value =>
-  `/icons.svg#${
+  `/design/icons.svg#${
```
viktor-yakubiv: Sorry, I missed this line.

package.json
```diff
@@ -74,14 +74,12 @@
   },
   "scripts": {
     "dev": "next dev",
-    "build:icons": "svgstore public/icons/* | svgo -i - -o public/icons.svg --disable=removeUselessDefs,cleanupIDs",
-    "prebuild": "run-p build:icons",
```
viktor-yakubiv: Still, removing this will slow down development of new features and make the releases harder. I would keep the process the same: move everything possible but continue development in the project. The next time move everything again.

Joozty:
What do you propose then?

Should I remove grey fill of all icons in design and add puzzle icon there or forget about it?

package.json
Joozty: As you can see in the commit message I don't say we shouldn't use it. We can add it back when needed but for now, we don't use it's not possible to comment these two lines in json file.

viktor-yakubiv:
If you want to use icons from the design, please update icons in the design with project icons, add my puzzle there too and merge first 2 commits.

However, please do not merge the last one because we may need it in the future and there is not need to look for it in the history.

PR #178, opened 2020-03-06
Fix activities

Joozty:
@viktor-yakubiv I know that you told me once that you don't like to have one global config for activities. But the current solution was for me very unintuitive. I spent half an hour by adding DOIs tab to menu and I don't lie. So in the first commit I moved everything into one file. All others commits fix small bugs. You can find more info in commit messages.

It's a proposal so I am up for anything. But it should be somehow improved. I wouldn't like if we let it be.

### DIFF
--- a/store/activities.js
+++ b/store/activities.js
@@ -2,28 +2,34 @@ const config = [
   {
     id: 'overview',
     path: 'overview',
+    icon: 'dashboard',
   },
   {
     id: 'content',
     path: 'content',
+    icon: 'document',
   },
   {
     id: 'depositDates',
     path: 'deposit-dates',
+    icon: 'calendar-check',
   },
   {
     id: 'plugins',
     path: 'plugins',
+    icon: 'puzzle',
+    children: [
+      {
+        id: 'plugins/discovery',
+        path: 'plugins/discovery',
+      },
+      {
+        id: 'plugins/recommender',
+        path: 'plugins/recommender',
+      },
+    ],
   },
-  {
-    id: 'plugins/discovery',
-    path: 'plugins/discovery',
-  },
-  {
-    id: 'plugins/recommender',
-    path: 'plugins/recommender',
-  },
-]
+].flatMap(x => [x, ...(x.children || [])])
 
 const configMap = new Map(config.map(item => [item.id, item]))
 
--- a/store/root.js
+++ b/store/root.js
@@ -80,6 +80,8 @@ class Root {
   @action changeActivity(url) {
     const activity =
       activities.find(({ path }) => url === path) || activities.get('overview')
+
+    if (this.activity?.path === url) return
     this.activity = activity
   }
 
--- a/components/application/activity-select/index.jsx
+++ b/components/application/activity-select/index.jsx
@@ -9,15 +9,7 @@ import activities from 'store/activities'
 const toUrl = value => (value === 'index' ? '.' : `./${value}`)
 const toIcon = value => `/design/icons.svg#${activities.get(value).icon}`
 
-const ActivitySelect = ({ children, value }) => (
-  <Drawer.List>
-    {React.Children.map(children, element =>
-      React.cloneElement(element, {
-        selected: element.props.value === value,
-      })
-    )}
-  </Drawer.List>
-)
+const ActivitySelect = ({ children }) => <Drawer.List>{children}</Drawer.List>
 
 const ActivitySelectOption = ({ children, value, selected, path }) => (
   <Drawer.Item className={styles[value]} href={toUrl(path)} active={selected}>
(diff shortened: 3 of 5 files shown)


texts/navigation.yml
```diff
@@ -4,7 +4,7 @@ items:
   index: Overview
   data: Data
   content: Content
-  deposit-dates: Deposit dates
+  depositDates: Deposit dates
```
viktor-yakubiv: Why did this change?

Joozty: that's a leftover. I was just testing something. Thanks

store/activities.js
```diff
@@ -2,31 +2,45 @@ const config = [
...
-  },
-  {
-    id: 'plugins/discovery',
-    path: 'plugins/discovery',
-  },
-  {
-    id: 'plugins/recommender',
-    path: 'plugins/recommender',
+    icon: 'puzzle',
+    subPaths: [
```
viktor-yakubiv: It may be better to call nested paths `children`.

```suggestion
    children: [
```

store/activities.js
```diff
@@ -2,31 +2,45 @@ const config = [
   {
     id: 'overview',
     path: 'overview',
+    icon: 'dashboard',
```
viktor-yakubiv: I do not agree. Store should know nothing about how it looks.

store/activities.js
```diff
@@ -2,31 +2,45 @@ const config = [
...
+    ],
   },
 ]
 
 const configMap = new Map(config.map(item => [item.id, item]))
 
 config.get = (...args) => configMap.get(...args)
+config.find = (...args) =>
+  config
+    .reduce((acc, curr) => {
```
viktor-yakubiv: Prefer `flatMap` here. I would even do `flatMap` once after creating the array.

store/activities.js
```diff
@@ -2,31 +2,45 @@ const config = [
...
+      acc.push(curr)
+      if (curr.subPaths) acc.push(...curr.subPaths)
```
viktor-yakubiv: Technically it's not correct. More specific routes should go first. In our case let's imagine that you match regular expression: `/plugins` would match `/plugins/discovery` so you always would get `plugins`.

```suggestion
      acc.push(...(curr.children ?? []), curr)
```

store/root.js
```diff
@@ -80,6 +80,8 @@ class Root {
   @action changeActivity(url) {
     const activity =
       activities.find(({ path }) => url === path) || activities.get('overview')
+
+    if (this.activity?.path === url) return
```
viktor-yakubiv: I would not add this. This is 404 and we have to handle this somehow. Maybe throwing would be better

Joozty: This is not 404. It's prevention for redirection loop. See [here](https://github.com/oacore/dashboard/blob/activities/pages/_app/index.jsx#L35)

store/activities.js
Joozty: Well, I was thinking about moving this config file somewhere else. For example to a new directory called `configs`?

store/activities.js
Joozty: Sorry, I don't get your example. If you use the regular expression the find method return you first match, which is `/plugins` and that's correct.

Joozty: Ah, I get it

Joozty: but anyway why would I use `/plugins` for matching /plugins/discovery?

store/activities.js
Joozty: That's really cool function. Thanks

store/activities.js
viktor-yakubiv: It's not about using plugins for matching `/plugins/discovery`. It's basically assumption that URL can be `/plugins/discovery#custom-settings` or `/plugins?viewAs=-1`. We should check only start of the URL and in this case the first option, i.e. `/plugins` would match.

store/root.js
viktor-yakubiv: When does the redirection loop happen?

Joozty: Basically we change dataProvider and activity on `routeChangeComplete` which triggers autorun action and this action changes Route. Route call `routeChangeComplete` and so on.

Basically I thought MobX fired autorun only when value changed. But apparently it fires it every time value is set.

store/activities.js
Joozty: And that's correct, isn't it? So if I have `/plugins/discovery#custom-settings`, `/plugins/recommender#custom-settings` and `/plugins#custom-settings` and I use only start url, i.e. `/plugins` which path should be matched? From what you are saying it should match `/plugins/discovery` and how do you know it's better then `/plugins/recommender`?

viktor-yakubiv: When you have a `test` and an actual `location`, and your tests are ordered
```js
[
  /^\/plugins/,
  /^\/plugins\/discovery/,
]
```

then testing a `/plugins/discovery#custom-settings` would match the first one, i.e. `/^\/plugins/` what is not correct.

If you use not `RegExp.prototype.test` but the `String.prototype.startsWith` the result would be the same.

---

Also, I am saying that we should rely on `startsWith` because there may be parameters in the end of the URL.

PR #179, opened 2020-03-09
Redirect to login when 401 occurs

Joozty:
@viktor-yakubiv What do you think? I think the best place for this is in apiRequest. Alternatively, we can create some decorator and handle it in store?

### DIFF
--- a/config.js
+++ b/config.js
@@ -3,10 +3,8 @@ export const API_URL =
     ? 'https://api.core.ac.uk/internal'
     : 'https://api.dev.core.ac.uk/internal'
 
-export const getLoginPage = (fullPath = false) =>
+export const getLoginPage = (path = window.location.origin) =>
   new URL(
-    `/test/login?continue=${
-      fullPath ? window.location.href : window.location.origin
-    }`,
+    `/test/login?continue=${encodeURIComponent(path)}`,
     process.env.NODE_ENV === 'production' ? undefined : API_URL
   ).toString()
--- a/config.js
+++ b/config.js
@@ -3,9 +3,6 @@ export const API_URL =
     ? 'https://api.core.ac.uk/internal'
     : 'https://api.dev.core.ac.uk/internal'
 
-export const LOGIN_URL =
-  process.env.NODE_ENV === 'production' ? '/login.html' : '/test/login'
-
 export const getLoginPage = (fullPath = false) =>
   new URL(
     `/test/login?continue=${
--- a/config.js
+++ b/config.js
@@ -3,11 +3,13 @@ export const API_URL =
     ? 'https://api.core.ac.uk/internal'
     : 'https://api.dev.core.ac.uk/internal'
 
-const ORIGIN = encodeURIComponent(
-  typeof window == 'object' ? window.location.origin : ''
-)
-
 export const LOGIN_URL =
-  process.env.NODE_ENV === 'production'
-    ? '/login.html'
-    : new URL(`/test/login?continue=${ORIGIN}`, API_URL).toString()
+  process.env.NODE_ENV === 'production' ? '/login.html' : '/test/login'
+
+export const getLoginPage = (fullPath = false) =>
+  new URL(
+    `/test/login?continue=${
+      fullPath ? window.location.href : window.location.origin
+    }`,
+    process.env.NODE_ENV === 'production' ? undefined : API_URL
+  ).toString()
--- a/pages/_app/index.jsx
+++ b/pages/_app/index.jsx
@@ -3,11 +3,12 @@ import NextApp from 'next/app'
 import { withRouter } from 'next/router'
 import { autorun } from 'mobx'
 
-import { getLoginPage } from '../../config'
-import Route from './route'
-
 import '@oacore/design/lib/index.css'
 
+import Route from './route'
+import { getLoginPage } from '../../config'
+
+import { UnauthorizedError } from 'api/errors'
 import logPageView from 'utils/analytics'
 import { initStore, GlobalProvider } from 'store'
 import Application from 'components/application'
@@ -76,8 +77,16 @@ class App extends NextApp {
     }
   }
 
-  redirectToLogin = () => {
-    window.location.replace(getLoginPage())
+  redirectToLogin = (path = false) => {
+    window.location.replace(getLoginPage(path))
+  }
+
+  handlePromiseRejection = event => {
+    if (event.reason instanceof UnauthorizedError) {
+      this.redirectToLogin(true)
+      // Don't report to console
+      event.preventDefault()
+    }
   }
 
   reflectStoreToRoute() {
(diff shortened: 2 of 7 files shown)


viktor-yakubiv:
I do not agree. API is a separate package and it should stay pure. All dirty things should be done in App.

Why the current implementation does not satisfy you?

Joozty:
Because I find it a security risk. (Very small but still...) If user is logged out for any reason they shouldn't see any data.

viktor-yakubiv:
So, you can add NetworkError catching to the App component and it it's an `UnauthorizedError` perform a redirect.

Joozty:
I wanted to avoid catching it in app component. It should be caught as soon as possible. We catch errors for endpoints differently. Some endpoints ignore API errors some propagate it. If we do it in APP component we will need to keep it in mind which results in duplicate code. This doesn't sound like a good software architecture for me.

viktor-yakubiv:
Duplication is not evil. Empty catch blocks are not good.

Mixing pure function with a side effect with no reason does not look like a good software architecture.

Joozty:
For me it's evil. 


Yes, I agree empty catch blocks are not good but we haven't agreed on error handling yet. I expect to either repeat an action or show some error message. For now I keep them empty but they should be extended.

> Mixing pure function with a side effect with no reason does not look like a good software architecture.
There is a reason. Protection.

viktor-yakubiv:
If you want to improve behaviour, you can map status codes to Error classes similar to following:
- `UnauthorizedError` for 401
- `PaymentRequiredError` for 402
- `ForbiddenError` for 403
- `NotFoundError` for 404
- `BadRequestError` for any ≥400, <500
- `InternalServerError` for ≥500

After that locally check for the type of error you want and propagate the rest.

store/deposit-dates.js
```diff
@@ -84,7 +85,7 @@ class DepositDates {
       const length = headers.get('Collection-Length')
       this.depositDatesCount = Number.parseInt(length, 10) || null
     } catch (e) {
-      if (e.status === 404) this.isExportDisabled = true
+      if (!(e instanceof NotFoundError)) this.isExportDisabled = true
```
viktor-yakubiv: The condition is opposite now. Should it be so?

Could you please call `error` always `error` and `event` always `event`. It may be confusing.

pages/_app/index.jsx
```diff
@@ -121,6 +131,10 @@ class App extends NextApp {
   componentWillUnmount() {
     const { router } = this.props
     router.events.off('routeChangeComplete', this.handleRouteChange)
+    window.removeEventListener(
+      'unhandledrejection',
+      this.handlePromiseRejection
+    )
```
viktor-yakubiv: Why do you need capture it separately? Does not `componentDidCatch` handle everything?

components/published-toggle/index.jsx
```diff
@@ -22,8 +24,10 @@ const PublishedToggle = ({
     try {
       await onVisibilityChanged(isVisible)
       toggleChecked(isVisible)
-    } catch (e) {
-      // something happened, don't update state
+    } catch (error) {
+      if (error instanceof UnauthorizedError) throw error
+
+      // something other happened, don't update state
```
viktor-yakubiv: Why cannot you pass errors above without any check?

config.js
```diff
@@ -3,11 +3,10 @@ export const API_URL =
...
-  process.env.NODE_ENV === 'production'
-    ? '/login.html'
-    : new URL(`/test/login?continue=${ORIGIN}`, API_URL).toString()
+export const getLoginPage = (fullPath = false) =>
+  new URL(
+    `/test/login?continue=${
+      fullPath ? window.location.href : window.location.origin
+    }`,
+    process.env.NODE_ENV === 'production' ? undefined : API_URL
+  ).toString()
```
viktor-yakubiv: It's better to either accept full path or nothing but never boolean.

```suggestion
export const getLoginPage = (path = ORIGIN_URL) =>
  new URL(`/test/login?continue=${encodeURIComponent(path)}`).toString()
```

pages/_app/index.jsx
```diff
@@ -134,6 +148,9 @@ class App extends NextApp {
 
       Sentry.captureException(error)
     })
+
+    if (error instanceof UnauthorizedError)
+      window.location.replace(getLoginPage(true))
```
viktor-yakubiv: Passing boolean value as a function parameter is a bad practice.

pages/_app/index.jsx
Joozty: Not it doesn't. See here: https://github.com/facebook/react/issues/11334#issuecomment-338656383

Actually I think it's useless to have check incomponentDidCatch. I cannot think about the use case for that.

config.js
Joozty: https://github.com/oacore/dashboard/pull/179/commits/c6e774c7222549a9b07c48310b57a38513c40c89

components/published-toggle/index.jsx
```diff
@@ -19,12 +19,8 @@ const PublishedToggle = ({
 
   const handleChange = useCallback(async event => {
     const isVisible = event.target.checked
-    try {
-      await onVisibilityChanged(isVisible)
-      toggleChecked(isVisible)
-    } catch (e) {
-      // something happened, don't update state
-    }
+    await onVisibilityChanged(isVisible)
```
viktor-yakubiv: What do you get back after actual changing `disable` property?

I imagine that:
1. you press a button
2. request is sent
3. store gets new entity in the response
4. store updates the entity
5. MobX triggers observers to update
6. React gets updated

config.js
```diff
@@ -3,11 +3,8 @@ export const API_URL =
...
 
-const ORIGIN = encodeURIComponent(
-  typeof window == 'object' ? window.location.origin : ''
-)
-
-export const LOGIN_URL =
-  process.env.NODE_ENV === 'production'
-    ? '/login.html'
-    : new URL(`/test/login?continue=${ORIGIN}`, API_URL).toString()
+export const getLoginPage = (path = window.location.origin) =>
```
viktor-yakubiv: If this is called on server, `window` is undefined. I would keep the following.

```suggestion
export const getLoginPage = (path = window?.location.origin) =>
```

pages/_app/index.jsx
```diff
@@ -3,11 +3,12 @@ import NextApp from 'next/app'
 import { withRouter } from 'next/router'
 import { autorun } from 'mobx'
 
-import { LOGIN_URL } from '../../config'
-import Route from './route'
-
 import '@oacore/design/lib/index.css'
 
+import Route from './route'
+import { getLoginPage } from '../../config'
```
viktor-yakubiv: Why this is not `from 'config'`?

Joozty: because of my stupid IDE. I need to tweak settings somehow.

config.js
Joozty: this won't help since window is not defined. Let's keep it as is till we run into this issue...

components/published-toggle/index.jsx
Joozty: That's what I wanted to do at first but it's not that easy to do it.

PR #180, opened 2020-03-10
DOIs tab

Joozty:


texts/navigation.yml
```diff
@@ -5,6 +5,7 @@ items:
   data: Data
   content: Content
   deposit-dates: Deposit dates
+  dois: DOIs
```
viktor-yakubiv: I would probably keep singular

store/activities.js
```diff
@@ -14,6 +14,11 @@ const config = [
     path: 'deposit-dates',
     icon: 'calendar-check',
   },
+  {
+    id: 'dois',
+    path: 'dois', // TODO: Find good icon
```
viktor-yakubiv: I would use either [barcode](https://iconify.design/icon-sets/mdi/barcode.html) or [QR-code](https://iconify.design/icon-sets/mdi/qrcode.html). Also, we may use [widgets](https://iconify.design/icon-sets/mdi/widgets.html) that means a registry in Windows but this is not as clear as previous ones.

pages/data-providers/[data-provider-id]/dois.jsx
```diff
@@ -0,0 +1,104 @@
...
+    },
+    {
+      id: 'publicReleaseDate',
+      display: 'Deposit date',
+      order: 'desc',
+      className: styles.depositDateColumn,
+      expandedSize: 1,
+      getter: v => dayjs(v.publicReleaseDate),
+      render: v => v.format('DD/MM/YYYY'),
+    },
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/dois.jsx]: Should be DOI only. I do not thing we need more columns but maybe @mcancellieri can help us here.

```suggestion
    {
      id: 'doi',
      display: 'DOI',
      className: styles.doiColumn,
      expandedSize: 1,
      getter: v => v.doi,
      render: v => v,
    },
```

components/application/logout.jsx
```diff
@@ -13,7 +13,7 @@ const Logout = ({ store }) => {
   const name = store.user.email
 
   return (
-    <AppBar.Item className={styles.container}>
+    <AppBar.Item>
```
viktor-yakubiv: Why do you change this? Just curious 🤔

pages/data-providers/[data-provider-id]/dois.css
```diff
@@ -0,0 +1,74 @@
...
+
+.card-one {
+  flex: 2 2 60%;
+  width: 100%;
+}
+
+.card-two {
+  flex: 1 1 30%;
+  width: 100%;
+}
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/dois.css]: Still do not know what to put there. The second one should stay being 'export'. The first one is a mystery at this moment 🙂

components/application/logout.jsx
Joozty: `container` class is not defined in logout.css. So I just simply removed it.

store/activities.js
Joozty: Perhaps I would go for barcode but it still doesn't explain the content of the tab to me.

viktor-yakubiv: Also, I thought about drawing something similar or getting actual favicon from doi.org. However, for this we might have to write them about asking a permission.

DOI.org does not have any branding resources and policy of using them on the website unfortunately and this process may take a long time.

@mcancellieri do you have any recommendation in the process? Should we write to DOI.org?

mcancellieri: i like the barcode one.
To be honest, I have never seen the favicon of doi.org used anywhere else so it will be as informative as the barcode, so I vote for the barcode.

pages/data-providers/[data-provider-id]/dois.css
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/dois.css]: Could you please remind me if we already decided? Or is it still a mystery?

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/dois.css]: Yes, it is. I've not been working on this yet, sorry.

pages/data-providers/[data-provider-id]/dois.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/dois.jsx]: The response contains these fields:

oai, repo_doi, crossref_doi,title, authors. I decided to show everything but oai and crossref_doi. Let me know if it needs to be changed.

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/dois.jsx]: I guess we need to show something like `repo_doi || crossref_doi`, ideally `crossref_doi && !repo_doi ? crossref_doi 🆕 : repo_doi`

Joozty:
Feel free to review if you have time. 

There are still missing two last pieces. First & overview card and renaming API fields to camelCase. I will continue once the backend is ready.

store/deposit-dates.js
```diff
@@ -20,7 +20,7 @@ class DepositDates {
 
   constructor(baseUrl) {
     const datesUrl = `${baseUrl}/public-release-dates`
-    this.publicReleaseDates = new Pages(`${baseUrl}/public-release-dates`)
+    this.publicReleaseDates = new Pages(datesUrl)
```
viktor-yakubiv: This may be fixed after rebase.

texts/doi/export.yml
```diff
@@ -0,0 +1,3 @@
+title: Export
+description: Download all your {{count}} DOI records as a CSV.
```
viktor-yakubiv: Perhaps the text should be different. 
```suggestion
description: Download {{count}} new DOI records as a CSV.
```

store/doi.js
```diff
@@ -0,0 +1,39 @@
+import { action, observable } from 'mobx'
+
+import { Pages } from './helpers/pages'
+
+import apiRequest from 'api'
+import { NotFoundError } from 'api/errors'
+
+const { API_URL } = process.env
+
+class DOI {
```
viktor-yakubiv: You will need to extend the `Store` after rebase.

store/doi.js
```diff
@@ -0,0 +1,39 @@
...
+    this.doiUrl = `${API_URL}${doiUrl}?accept=text/csv`
+
+    this.loadDOICount()
+  }
+
+  @action
+  loadDOICount = async () => {
+    try {
+      this.isExportDisabled = false
+      const { headers } = await apiRequest(this.doiUrl, { method: 'HEAD' })
```
viktor-yakubiv: After extending Store

```suggestion
      const { headers } = await this.request(this.doiUrl, { method: 'HEAD' })
```

pages/data-providers/[data-provider-id]/doi.jsx
```diff
@@ -0,0 +1,75 @@
...
+        title="Browse DOI records"
+        pages={store.doi.doiRecords}
+        searchable
+      >
+        <Table.Column
+          id="repo_doi"
+          display="DOI"
+          order=""
+          getter={({ crossrefDoi, repoDoi }) =>
+            crossrefDoi && !repoDoi ? `${crossrefDoi} 🆕` : repoDoi
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/doi.jsx]: It would be nice to have an icon instead 🆕. It's hard to append an icon to the end of the text in the design. However, here it's not a big deal apart from adding [the icon](https://iconify.design/icon-sets/mdi/new-box.html) into the library.

pages/data-providers/[data-provider-id]/doi.jsx
```diff
@@ -0,0 +1,75 @@
...
+    <Card className={styles.doiTable} tag="section">
+      <Table
+        key={store.dataProvider}
+        title="Browse DOI records"
+        pages={store.doi.doiRecords}
+        searchable
+      >
+        <Table.Column
+          id="repo_doi"
+          display="DOI"
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/doi.jsx]: I would place after the title. Is there any reason to have such order of columns you have here?

pages/data-providers/[data-provider-id]/doi.jsx
```diff
@@ -0,0 +1,75 @@
...
+          order=""
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/doi.jsx]:
```suggestion
          order="any"
```
Sounds more clear when you read it.

pages/data-providers/[data-provider-id]/doi.jsx
```diff
@@ -0,0 +1,75 @@
...
+          className={styles.authorsColumn}
+          getter={v => v.authors && v.authors.map(a => a.name).join(' ')}
+        />
+      </Table>
+    </Card>
+  </Tag>
+)
+
+export default withGlobalStore(DepositDates)
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/doi.jsx]: Why cannot you just pass the `store` as a prop from the `App` and do not use this context at all?

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/doi.jsx]: this decorator makes class observe, i.e. if the store would be taken from APP it wouldn't retrigger rerender when `store.depositDates.depositDatesCount`  changes. So either you add `@observe` decorator or use this function. Both approaches are almost identical.

pages/data-providers/[data-provider-id]/doi.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/doi.jsx]: Not really it's just in other tables we always show ID as a first column. So I just wanted to keep consistency.

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/doi.jsx]: Sorry, I missed that we do not have OAI here. However, I think the repository manager would be interested to match OAI-DOI to avoid title/author matching, so if it's in the exported file, it may be needed in here too. @mcancellieri?

viktor-yakubiv:
@Joozty I added icons, merged and rebased our work together. I need to fix the statistics endpoint call and NaN in the overview card.

Feel free to review commit messages at this stage and the code if you have time. Thank you! 🙂

pages/data-providers/[data-provider-id]/doi.jsx
```diff
@@ -6,14 +6,33 @@ import NumericValue from 'components/numeric-value'
...
-import { Card } from 'design'
+import { Card, Icon } from 'design'
 
 // TODO: Remove once cards are in @oacore/design
 // eslint-disable-next-line
 import styles from './doi.css'
 
 const formatPercent = (number, precision = 2) => `${number.toFixed(precision)}%`
 
+const formatDOI = entity => {
```
Comment (Joozty) [inline pages/data-providers/[data-provider-id]/doi.jsx]: Not sure if you know but you there is a better way of doing this. By extending Table.Column. But basically it's the same. We maybe just want to get rid of `getter` in the future so this approach would be better. It's up to you I don't have preference here.

```jsx
class DOIColumn extends Table.Column {
	render() {
      const { context: {crossrefDoi, repoDoi: originDoi} } = this.props
	  if (crossrefDoi && !originDoi) {
    	return (
	      <>
	        {crossrefDoi}{' '}
	        <Icon
	          className={styles.newIcon}
	          src="/design/icons.svg#new"
	          alt="New"
	        />
	      </>
	    )
	  }
	  return originDoi
	}
}
```

pages/data-providers/[data-provider-id]/overview.jsx
```diff
@@ -129,6 +84,17 @@ const DepositingCard = ({ chartData, complianceLevel }) => {
   )
 }
 
+const DOICard = ({ value }) => (
+  <Card>
+    <h2>DOI coverage</h2>
+    <PerformanceChart value={value} increase={2} caption="DOI" />
+    {2 && <p>We can enrich you DOI coverage by {2}%.</p>}
```
Comment (Joozty) [inline pages/data-providers/[data-provider-id]/overview.jsx]: Maybe it's worth putting this into texts?

pages/data-providers/[data-provider-id]/doi.jsx
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/doi.jsx]: I would leave this on you in the future refactoring if you do not insist. 

P.S. React guys would not like extending, do they? 🙂

components/numeric-value/index.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+
+import { Numeral } from 'design'
+
+const formatNumber = (n, { locale = 'en-GB' } = {}) =>
+  n > 1000
+    ? `${(Math.floor(n / 10) / 100).toLocaleString(locale)}\xa0K`
```
Joozty: Would you mind putting a comment here explaining that magic `\xa0K` meaning because honestly, I don't have any idea?

viktor-yakubiv: Yes, I will do it when I find it myself 🙂 It's either space or special mathematical char.

pages/data-providers/[data-provider-id]/doi.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/doi.jsx]: I don't really know but for example, the whole material-ui library is built on extended components if I remember correctly. To be honest I am more functional guy and don't like classes in JS.

pages/data-providers/[data-provider-id]/overview.jsx
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/overview.jsx]: Yes, I will do so

viktor-yakubiv:
@Joozty thank you for the review. I fixed and improved some things. I have to move strings to texts now.

I really need your help at this moment. I often get redirect hell. I am afraid this is because App is an observer but I am not sure. You can reproduce the bug if you go to DOI tab and press 'back' button.

Also, I am struggling with processing loading state everywhere.

store/doi.js
```diff
@@ -19,7 +19,8 @@ class DOI extends Store {
         : null
 
     if (lag != null && lag < 0) {
-      console.error('DOI enrichment count is less then 0!', lag)
+      if (process.env.NODE_ENV === 'development')
+        console.error('DOI enrichment count is less then 0!', lag)
```
Joozty: Maybe it would be worth to trowing it and log it into [Sentry](https://github.com/oacore/dashboard/blob/master/pages/_app/index.jsx#L92)?

Joozty:
The loading state is tricky. What about storing object into `enrichmentSize` or some decorator or something. I don't really know to be honest.

```
{
   "state": "loading|loaded",
   "value": 0
}
```

I checked your PR and it looks good to me. 👍 If there are some bugs I would propose to fix it in new PRs.

store/doi.js
viktor-yakubiv: How Sentry would work on development? Will it log to console?

I would like to log to console during development and report to Sentry if production.

viktor-yakubiv:
@Joozty what bugs exactly do you mean?

viktor-yakubiv:
Since there are no bugs, thankfully to #221, I am going to rebase this branch and prepare it to the final review and release.

store/doi.js
Joozty: AFAIK it logs it always. Basically you will see something like `Uncaught in promise Error: <error-message> and stacktrace`. I didn't test it. So maybe it doesn't get caught at all by the even listener.

viktor-yakubiv:
@Joozty do you have any ideas why during the data loading numbers weirdly change?

About the loading state reporting, I have a couple of ideas:
1. Extend build-ins adding a loading property. Unfortunately, we cannot extend `null` and the Number is tricky too.
2. Introduce scopes similarly to [swr](https://swr.now.sh). Basically, their idea in this is quite good and we can either use the framework (warning, it's under 1.0) or develop something similar

Joozty:
@viktor-yakubiv What numbers?

viktor-yakubiv:
Numbers on the charts. It may be an impact of data checking for null and zeros but needs an investigation.

components/enrichment-chart/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+          formatter={value => Number(value).toLocaleString('en-GB')}
+          fill="#fff"
+          offset={16}
+        />
+      </Bar>
+    </BarChart>
+  </ResponsiveContainer>
+)
+
+const EnrichmentChart = ({ value, caption, increase }) => {
```
Joozty: Can you please put here check for a null value. Otherwise, if you go directly to https://dashboard-git-dois.oacore.now.sh/data-providers/27/doi it will fail because of some sort of recursion problem in recharts library. (They expect that props are correct)

```
const EnrichmentChart = ({ value, caption, increase }) => {
  if (!value) return null
  const data = [
    increase && {
      caption: 'CORE',
      value: value + increase,
      fill: 'var(--primary)',
    },
    {
      caption,
      value,
      fill: increase ? 'var(--gray-700)' : 'var(--primary)',
    },
  ].filter(truthy => truthy)

  return <VerticalBarChart data={data} />

```

Joozty:
Apart from this [issue](https://github.com/oacore/dashboard/pull/180#discussion_r401004609), I don't see anything wrong with numbers. Can you please share video/screenshot or anything which could help me to help you here (what a sentence 😄)?

pages/data-providers/[data-provider-id]/doi.jsx
```diff
@@ -0,0 +1,143 @@
...
+          alt="New"
+        />
+      </>
+    )
+  }
+
+  return originDoi
+}
+
+// const useLoading = (func, deps = []) =>
```
Comment (Joozty) [inline pages/data-providers/[data-provider-id]/doi.jsx:38]: Could you please remove this commented code?

pages/data-providers/[data-provider-id]/doi.jsx
```diff
@@ -0,0 +1,143 @@
...
+//     const fetched = deps.reduce((done, value) => done && value != null, true)
+//     if (!fetched) return null
+//     return func()
+//   }, deps)
+
+const useDefault = (value, substitute = null) =>
+  value == null ||
+  value === Infinity ||
+  value === -Infinity ||
+  Number.isNaN(value)
```
Comment (Joozty) [inline pages/data-providers/[data-provider-id]/doi.jsx:49]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite This function will cover +-Infinity and isNaN cases

PR #184, opened 2020-03-12
Add loading login form instead blank screen

Joozty:
Adds a login page with a loading spinner to avoid the blank screen during data retrieving.

### DIFF
--- a/pages/_app/index.css
+++ b/pages/_app/index.css
@@ -3,3 +3,8 @@ body,
 :global(#__next) {
   height: 100%;
 }
+
+.login-iframe {
+  width: 100%;
+  height: 100%;
+}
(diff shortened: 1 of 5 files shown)


Joozty:
> Do you restore previous view if loading gets an error?

Page gets reloaded if the wrong credentials are typed in.

Also, the form is empty when showed in iframe. There is nothing we can do about it. (we don't know username) but it's still better than a blank screen.

viktor-yakubiv:
Could we please discuss #185?

viktor-yakubiv:
I get it, you just fake it, right? You do redirect anyway but to avoid blank screen you render login. Is this correct?

Joozty:
> I get it, you just fake it, right? You do redirect anyway but to avoid blank screen you render login. Is this correct?

Yes, correct and I handled all cases of the blank screen.

PR #186, opened 2020-03-13
iframe log in

Joozty:
I didn't want to create pages/login.jsx because it could cause a circular dependency. Or at least it would cause a flickering screen.

My idea is the following:

1. when a page is being loaded we fire request for a user to see if the user is logged into it. While this happens animation is shown. (here may happen that email input is empty)
3. if user is logged in we simply init store and render children
4. if the user is not logged in parent send an event to iframe to hide animation
user can log in
5. when user submits form then animation is shown
6. when login_check is successful iframe sent an event to parent to requery /user otherwise repeat the process
7. Logout is merged with login - we show info message that user is logged out and they can login again.

We could possibly have /login page. it would be almost the same. Just there would be two redirects (one from app to login and then back) in case user is not logged in.

Actually, when I think about it may be even more complicated. I find easier to keep everything in one place.


public/login.js
```diff
@@ -1,15 +1,36 @@
...
+  const formData = new FormData(event.target)
+  const data = new URLSearchParams(formData)
+  fetch('https://api.dev.core.ac.uk/login_check', {
+    method: 'POST',
+    body: data,
+    credentials: 'include',
+  }).finally(() => {
+    if (window.location !== window.parent.location) {
+      // The page is in an iframe
+      window.parent.postMessage('log-in-processing', '*') // TODO: core.ac.uk for production
```
viktor-yakubiv: Use `window.location.host`, it will save you.

public/login.js
```diff
@@ -23,4 +44,6 @@ window.addEventListener('DOMContentLoaded', () => {
   form.addEventListener('submit', login)
 })
 
-window.addEventListener('login-in-processing', addLoadingAnimation, false)
+window.addEventListener('login-in-processing', addLoadingAnimation)
+window.addEventListener('login-in-finished', removeLoadingAnimation)
```
viktor-yakubiv: I would handle this in iframe only and to not pass above. What the reason to pass it to the parent app?

Joozty:
I didn't want to create pages/login.jsx because it could cause a circular dependency. Or at least it would cause a flickering screen. 

My idea is the following:

1. When a page is being loaded we fire request for a user to see if the user is logged it. While this happens animation is shown. (here may happen that email input is empty)
2. if user is logged in we simply init store and render children
3. if the user is not logged in parent send an event to iframe to hide animation
4. user can log in
5. when user submits form then animation is shown
6. when login_check is successful iframe sent an event to parent to requery /user otherwise repeat the process


We could possibly have /login page. it would be almost the same. Just there would be two redirects (one from app to login and then back) in case user is not logged in.


Actually, when I think about it may be even more complicated. I find easier to keep everything in one place.

public/login.js
Joozty: I tried to explain in it description comment

Joozty:
@viktor-yakubiv Feel free to do early review. I am blocked by [backend](https://kmi-ou.atlassian.net/browse/CORE-3049) now. I will continue once backend support is implemented.

components/application/logout.jsx
```diff
@@ -6,10 +6,11 @@ import styles from './logout.css'
 import { withGlobalStore } from 'store'
 
 const LOGOUT_URL = `https://api.dev.core.ac.uk/logout?continue=${encodeURIComponent(
-  typeof window !== 'undefined' ? `${window.location.origin}/logout.html` : ''
+  typeof window !== 'undefined' ? `${window.location.origin}/?logout` : ''
```
viktor-yakubiv: Assuming `typeof` always returns a string by definition you [can use `!=` here](https://eslint.org/docs/rules/eqeqeq#smart).

pages/_app/index.jsx
```diff
@@ -178,18 +200,45 @@ class App extends NextApp {
     return null
   }
 
+  Login = () => {
```
viktor-yakubiv: This looks weird, if you want this to be a component move this out of the _App_ and pass props, an `onLogIn` callback or whatever you need.

pages/_app/index.jsx
```diff
@@ -178,18 +200,45 @@ class App extends NextApp {
...
+    return (
+      <iframe
+        ref={this.loginIframeRef}
+        title="Login Form"
+        src={`/login.html?${param}`}
+        className={styles.loginIframe}
+        onLoad={() => {
+          const { isAuthorized, loginInProcess } = this.state
+          if (!isAuthorized && !loginInProcess) {
+            this.loginIframeRef.current.contentWindow.postMessage(
```
viktor-yakubiv: If you use onLoad iframe event and subscribe from the parent component to it passing a function seems there is no need in posting a message to a window object.

However, if there is a need in that, the `login.js` should post it to own `window`.

pages/_app/index.jsx
```diff
@@ -178,18 +200,45 @@ class App extends NextApp {
...
-          src="/login.html?loading"
-          className={styles.loginIframe}
-        />
+        <>
+          <Application
+            dataProvider={undefined}
+            activity="overview"
+            onClick={this.handleNavigation}
+          />
+          <Login />
```
viktor-yakubiv: Why you did not pass this as an Application's child?

public/login.css
```diff
@@ -55,13 +55,19 @@ button:hover {
...
+}
+
 .message:empty {
   display: none;
 }
 
 /* stylelint-disable-next-line selector-max-id */
 #login-form {
   position: relative;
+  background-color: #fff;
```
viktor-yakubiv: If you render an iframe inside of something you can adjust inputs to get full window width and and render the iframe inside of a `Card`. I would look better.

public/login.js
```diff
@@ -1,26 +1,54 @@
...
+}
+
 window.addEventListener('DOMContentLoaded', () => {
   const urlParams = new URLSearchParams(window.location.search)
   if (urlParams.has('reason')) {
     const message = document.getElementById('error-message')
     message.innerHTML = 'The username or password you entered is incorrect.'
   }
 
+  if (urlParams.has('logout')) {
```
viktor-yakubiv: I would modify it to `reason=wrong-credentials` and `reason=logout`.

Doing so you can:
1. Have a single message container
2. Switch the message text and class name based on the `reason`.

pages/_app/index.jsx
Joozty: I saw it used quite a lot in react projects. The reason why I used it here is that it is granted that iframe ref is assigned before componentDidMount is called. If it is extracted ref may not be assigned.

components/application/logout.jsx
Joozty: I always use strict comparison. Otherwise, my IDE highlights it and I don't want that.

pages/_app/index.jsx
Joozty: onLoad event is there just for a case when /user endpoint responds faster than iframe is loaded. 

> However, if there is a need in that, the login.js should post it to own window.

I don't get it. Can you provide some sample code?

components/application/logout.jsx
viktor-yakubiv: If you use JetBrains' IDE, it always highlight too much but it should prefer ESLint config in cases such as this one.

IDE is intelligent by definition but you should keep in mind that it's done by people for people and cannot be more smart than people themselves. So, it's pretty fine that IDE highlights something that should not be highlighted.

pages/_app/index.jsx
viktor-yakubiv: There is no need in sample. If you post a message from the iframe in the script that is executed in the iframe, the parent should be able to subscribe and get messages.

Also, you can post a message directly to parent using `window.parent.postMessage`.

viktor-yakubiv: I could not find a case when the [window.parent](https://developer.mozilla.org/en-US/docs/Web/API/Window/parent) is not set but I am sure there is a way to disable it using some [noopener, noreferer policy](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types). However, I am sure in our case everything should be fine because we use the same origin.

By the way, I found you may use [window.frameOrigin](https://developer.mozilla.org/en-US/docs/Web/API/Window/frameElement) and may even try to fire [custom event](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) onto it, e.g. `login` or `logout`.

components/application/logout.jsx
Joozty: I don't see anything wrong with using `!==`. The eslint doesn't enforce using `!=` with typeof. Both ways are correct it's just a matter of taste unless

pages/_app/index.jsx
Joozty: The parent [subscribes](https://github.com/oacore/dashboard/blob/login/pages/_app/index.jsx) to those events.

Also, I use window.parent.postMessage for sending a message to the parent.

The reason why it is like that is that iframe is loaded in the `loading` state, i.e. user sees animation. _app calls /user endpoint in case a user is not logged in the animation is hidden. Onload event handles the case when iframe haven't subscribed to the post message event yet.

PS: line highlighting here in GitHub doesn't work for me. So links will just redirect you to the files.

pages/_app/index.jsx
Joozty: I think I answered it here https://github.com/oacore/dashboard/pull/186#discussion_r392923262

pages/_app/index.jsx
viktor-yakubiv: Saying that I was assuming that you move ref into the Login component and handle all login process inside exposing only simple events.

pages/_app/index.jsx
viktor-yakubiv: It's not correct to use the login iframe just because to want to show a loading state instead the blank screen. It's 2 different problems and there are 2 different solutions.

Login should solve only its own one.

Joozty: Well, I could extract animation to parent window. I sit that what you propose?


PS: To be honest I see things differently. We've got a problem to solve and there are unlimited ways how to solve it. I don't say something is not correct. It's just a different approach to solve the problem.

pages/_app/index.jsx
viktor-yakubiv: I still do not understand why this is not a child.

pages/_app/index.jsx
viktor-yakubiv: This, this is what I propose. I agree with P.S., we see different ways of doing this. I am not saying your way is wrong.

viktor-yakubiv:
I propose to postpone this PR in favour of #188

Joozty:
I don't want to postpone it. There are two of my PRs already postponed. It's hard to keep track and return back to it. :(

pages/_app/index.jsx
Joozty: What do you mean? Like login should be a new component?

viktor-yakubiv: The _Login_ is a ~new~ component. You render it along with the _Application_ as a sibling. I am just curious why do not you render it as an Application's child?

pages/_app/index.jsx
Joozty: Ok, I will extract it.

The reason why I am saying it is that sometimes I've got a feeling that you assign me a task where you have a clear idea of how to do it. I do it differently. (Sometimes completely different 😄 sometimes only a little bit). Then you check the code and write different things. I like constructive comments but some of them are too opinionated in my opinion and you kind of insist on them. I consider it wasting time because now two of us work on the same task. You want me to implement your idea, which means I develop something and you review and the cycle repeats. Maybe it would be sometimes better if you take your task. I know you have plenty of other things to do but still. Please don't take it too closely. 🙂

viktor-yakubiv: You have a good point. I do not mind merging it as it is but I see there is a way to do it better. Your work is really appreciated.

Feel free to point me into the opinionated comments because I may not always notice that.

pages/_app/index.jsx
Joozty: Oh yes. The idea behind it is that I wanted to have blurred submenu too. (at least for logout). Also now the log in form is centered both vertically and horizontally. It could be achieved as a child too I guess.

viktor-yakubiv:
I tried to test this but I cannot login. Technically I can login, the cookie is set but the `/user` endpoint returns 401. @Joozty and @mcancellieri do you experience the same issue?

public/login.html
```diff
@@ -12,6 +12,7 @@
 
 <form id="login-form" method="post" class="">
   <h1>Login</h1>
+  <div id="info-message" class="message message-info"></div>
   <div id="error-message" class="message"></div>
```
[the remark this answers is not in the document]
Joozty: https://github.com/oacore/dashboard/pull/186/commits/08044326309c37832766b62cc40d95d414d6c7a9

mcancellieri:
I can't log in either, but it's a CORS issue because of the deployment URL.

viktor-yakubiv:
It's weird that I can log in on the [another page](https://dashboard.oacore.now.sh/) on the same domain area without any problems.

It's not a CORS problem because there is no CORS request. The cookie should be set at least because the requests goes to `api.core.ac.uk` and the application there sets the cookie for itself, i.e. `api.core.ac.uk`. This does not happen and I wonder if we forget to add some bits into fetch call. @Joozty what do you think?

Joozty:
Sorry, I forgot to mention that I haven't solved dev env yet. I fixed the issue you experienced. (You won't be able to log in when you run this code locally)

This is actually something I wanted to ask you. I don't know how to solve dev environment, i.e. which API url should be used. It's not possible to pass env variable to login.js file. I can only check window.location.origin and determine env out of that or to pass search param to login iframe like `login.html?isDev=true`

viktor-yakubiv:
I would probably pass `env` parameter that can be equally:
- empty or `prod` – `api.core.ac.uk`
- `dev` – `api.dev.core.ac.uk`

At some point we can adjust it to:
- empty or `prod` – `api-prod.core.ac.uk`
- `dev` – `api-dev.core.ac.uk`

I found we already have this convention and I would like to adapt it for the API and the Dashboard. @mcancellieri I would be good for you because `api-dev.core.ac.uk` would have a load balancer as well as `api.core.ac.uk`.

---

There is no correct way to do this. First, I was thinking to pass a variable as a parameter, i.e `login.html?api=api.core.ac.uk` but the problem with that is the URL itself can be modified before passing to something like `example.com?source=api.core.ac.uk`.

Joozty:
What if we move the login process to API itself? It's kind of related to API anyway. The parameter is not safe enough I agree. Alternatively, we could move this to the new repo, process it with some webpack/gulp/parcel and deploy it under login.core.ac.uk or smth like that.

viktor-yakubiv:
I support the second option but I do not agree with the first one. API should have auth but should not provide tokens. Although, I may be wrong. I will google something.

Joozty:
I just don't know if new repo for a hundred lines of code is a good move. Try to google something and please get back here.

viktor-yakubiv:
It's not about a repo with 100 lines. It's about moving identity provider to a separate design entity, or a service if you prefer to say that.

A picture from the Internet to show it:

Identity provider

It's not really good example but is clearly defines an Identity Provider as a separate entity which provides a token to a user and validates the token on request from the API. This is what would be nice to have and host somewhere under `id.core.ac.uk`. We may have this service on whatever language/framework but host it together to speed up the verification process.

Joozty:
It kind of is because it is a new service to take care of. I can imagine having it like that but in that case, it would make sense to separate API too and that's not gonna happen. If we have there more stuff, like user account info, etc. It would make sense to separate it. 

I don't know. :/ Maybe we should move this discussion to Slack and ask for other's opinions.

viktor-yakubiv:
@Joozty did you solve development environment?

I thing that if we create a variable called `IDENTITY_PROVIDER_URL` and pass the url into iframe it should be fine. It's simple, it's not 100% secure but we probably not need such level of security now. It would be fine if we use cookie authorisation because foreign identity provider cannot associate cookie with `core.ac.uk`. The only way to bypass this is to pass `javascript:sendCredentials()`

viktor-yakubiv:
I've just tested it. It's quite safe.

Browser throws an error in `cors` mode no matter you use Fetch API or XMLHttpRequest. The error looks like:

> Access to XMLHttpRequest at 'javascript:alert("hello");' from origin has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https.

In `no-cors` mode another kind of error happens.

It's safe to pass `IDENTITY_PROVIDER_URL` to login page. Let's go this direction.

Joozty:
Can you please tell me how did you test? It's not clear from the error message what did you do.

viktor-yakubiv:
I executed in the Chrome's console:

```js
fetch('javascript:alert("hello")')

fetch('javascript:alert("hello")', { mode: 'no-cors' })

let xhr = new XMLHttpRequest()
xhr.open('GET', 'javascript:alert("hello")')
xhr.send()
```

Joozty:
That doesn't make sense. Why would you do it? You provided an invalid URL to fetch.

Anyway, you won't be able to fetch anything except core.ac.uk and self domain because of CSP.

viktor-yakubiv:
It does because the following example would work as expected:
```html
<form action="javascript:alert('hello')">
  <input type="submit" value="Click me!">
</form>
```

public/login.js
```diff
@@ -1,26 +1,79 @@
...
+    body: data,
+    credentials: 'include',
+  })
+    .then(r => {
+      if (r.status === 401) showWrongCredentialsMessage()
+    })
+    .finally(() => {
+      if (window.location !== window.parent.location) {
+        // The page is in an iframe
+        window.parent.postMessage('login-processing', window.location.origin)
```
viktor-yakubiv: I see no problem if you post message to yourself. It would be just ignored, no need in condition here.

public/login.js
```diff
@@ -1,26 +1,79 @@
...
+        break
+      case 'logout':
+        showLogoutMessage()
+        break
+      default:
+    }
   }
 
-  if (urlParams.has('loading')) addLoadingAnimation()
+  if (urlParams.get('loading') === 'true') addLoadingAnimation()
```
viktor-yakubiv: It looks weird, prefer checking for existence of the parameter and pass it properly from the parent.

pages/_app/index.jsx
```diff
@@ -178,18 +200,51 @@ class App extends NextApp {
     return null
   }
 
+  Login = () => {
+    const { router } = this.props
+
+    const search = new URLSearchParams({
+      reason: router.query.reason === 'logout' ? 'logout' : '',
+      loading: router.query.reason !== 'logout',
+      identity_provider_url: API_URL.replace('internal', ''),
```
viktor-yakubiv: It's better to create a separate configuration property called `IDENTITY_PROVIDER_URL` and make the replacement there.

Joozty:
@viktor-yakubiv do you want to hold this one too? Can it be included in the release?

public/login.js
Joozty: Yep, you are right but it doesn't spare any line of code because I decided to support login from /login.html too. So there still will be an if statement.

viktor-yakubiv:
Yes, I think it can

pages/_app/index.jsx
Joozty: I was thinking it and ended up with this approach. It's used only once. If we use this URL somewhere else we can create configuration property for it.

public/login.js
Joozty: https://github.com/oacore/dashboard/pull/186/commits/1224a9bd1e665960160b0b6de07c3ad028a2833d

PR #189, opened 2020-03-16
Fix from param

Joozty:
Fix for https://kmi-ou.atlassian.net/browse/CORE-3054

### DIFF
--- a/store/helpers/pages/page.js
+++ b/store/helpers/pages/page.js
@@ -11,7 +11,7 @@ class Page {
   }
 
   get isLast() {
-    return this.data.length < this.maxSize
+    return this.data.length === 0
   }
 }
 
--- a/store/helpers/pages/pages.js
+++ b/store/helpers/pages/pages.js
@@ -48,7 +48,7 @@ class Pages {
     const order = getOrder(this.#columnOrder)
 
     const params = {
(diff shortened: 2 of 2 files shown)


store/helpers/pages/page.js
```diff
@@ -11,7 +11,7 @@ class Page {
   }
 
   get isLast() {
-    return this.data.length < this.maxSize
+    return this.data.length === 0
```
viktor-yakubiv: How is this affected?

mcancellieri:
It's a bug in production, is fine to wait as our production is not really production yet :)

store/helpers/pages/page.js
Joozty: You can find more in the task description. Basically when 100 records are requested API can return only 10 but it doesn't mean there are no more records. The last page is detected when empty array is sent.

viktor-yakubiv: Yes but this is not correct behaviour, is it? @valerapro could you please look at this?

Joozty: It was fixed like that. I just did the frontend part.  Like I don't like how it works but I understand it's hard to process it on the backend. So two lines of changes on the frontend are sufficient for me.

PR #190, opened 2020-03-16
Stream exported csv file

Joozty:
@viktor-yakubiv Could you please help me with styling clicked export button. Now it gets red color and red underline. Should it stay white without underline?

### DIFF
--- a/pages/data-providers/[data-provider-id]/deposit-dates.jsx
+++ b/pages/data-providers/[data-provider-id]/deposit-dates.jsx
@@ -1,6 +1,5 @@
 import React from 'react'
 import dayjs from 'dayjs'
-import { classNames } from '@oacore/design/lib/utils'
 
 import { withGlobalStore } from 'store'
 import Markdown from 'components/markdown'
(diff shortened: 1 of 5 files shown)


viktor-yakubiv:
It should but this is a problem of the design library. I would keep it as is and fix it in the design later, it's not a big problem.

Is there a way to preview disabled button without running project locally?

pages/data-providers/[data-provider-id]/deposit-dates.jsx
```diff
@@ -108,16 +110,25 @@ const DepositDates = ({
...
-        variant="contained"
-        onClick={() => store.depositDates.exportCsv()}
-        disabled={
-          store.depositDates.isExportInProgress ||
-          store.depositDates.isExportDisabled
-        }
+      <span
+        className={classNames.use({
+          [styles.exportButtonDisabled]: store.depositDates.isExportDisabled,
+        })}
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-dates.jsx]: Is there any problem with passing a class name into the _Button_ directly?

pages/data-providers/[data-provider-id]/deposit-dates.jsx
```diff
@@ -8,6 +9,7 @@ import TimeLagChart from 'components/time-lag-chart'
 import { Button, Card, Label } from 'design'
 import * as texts from 'texts/depositing'
 import DocumentLink from 'components/document-link'
+import { API_URL } from 'config'
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-dates.jsx]: I would like to avoid URL concatenation here. Can you add a pre-calculated property into the store and use it from there?

Joozty:
I am afraid you will need to run this project locally to see disabled button. Or you can try every repository and maybe you find empty one.

pages/data-providers/[data-provider-id]/deposit-dates.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-dates.jsx]: Yes, the cursor: not allowed doesn't work. pointer-events takes precedence over it.

pages/data-providers/[data-provider-id]/deposit-dates.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-dates.jsx]: https://github.com/oacore/dashboard/pull/190/commits/049212e282c84a297c15575d02cc4ab911648a73

viktor-yakubiv:
@Joozty I removed extra span, class and resolved the commit. Could you please confirm the business logic has not changed?

PR #191, opened 2020-03-17
Simplify apiRequest interface

viktor-yakubiv:
Simplifies `apiRequest` function to only 2 parameters similarly to `fetch` interface.

Refactors the `api` module. Removes `CancelablePromise` usage in favour of native Fetch API.

Change invalidatePreviousRequests to use native AbortController's signal to append to the fetch request.

Removes redundant `CancelablePromise`.

---

@Joozty my goal was to simplify `apiRequest` interface but I ended up rewriting a lot of stuff that is highly coupled. Honestly and unfortunately it's refactoring for the refactoring and I am not really proud of but I hope this may be helpful.


api/index.js
```diff
@@ -1,89 +1,100 @@
...
+  // user is probably offline or CORS failed
+  throw new NetworkError(
+    `Request for ${response?.url} failed. Response: ${response?.status}, ${message}`
+  )
+}
+
+const executeRequest = ({ url, ...options }) => {
+  fetch(url, options)
+    .then(processStatus)
+    .then(processBody)
```
Joozty: I think you are missing here `.catch(processError)`

api/index.js
```diff
@@ -1,89 +1,100 @@
...
+  if (error instanceof NetworkError) throw error
+
+  // user is probably offline or CORS failed
+  throw new NetworkError(
+    `Request for ${response?.url} failed. Response: ${response?.status}, ${message}`
+  )
+}
+
+const executeRequest = ({ url, ...options }) => {
+  fetch(url, options)
```
Joozty: you are missing here return :)

viktor-yakubiv: It's debug. Thank you

api/index.js
viktor-yakubiv: I do it for a reason. I cannot and do not want to pass `request` object to error processing. So if I remove the second part (converting any error to the `NewtorkError`), there is no need in it at all.

However, you put it there for a reason. If you insist I will put it back. Why did you convert any error to `NetworkError`?

api/index.js
```diff
@@ -1,89 +1,100 @@
...
+
+const prepareMethod = ({ method = 'GET' }) => method.toUpperCase()
+
+const prepareBody = ({ method, body }) => {
+  if (!['GET', 'HEAD'].includes(method))
+    return typeof body == 'object' ? JSON.stringify(body) : body
+  return null
+}
+
+const prepareHeaders = ({ headers: customHeaders, body }) => {
```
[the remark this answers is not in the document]
viktor-yakubiv: It will not work so. `:` means renaming.

I see no reason to assign a value. `{ ...undefined }` would not throw.

api/index.js
```diff
@@ -1,89 +1,100 @@
...
+
+const executeRequest = ({ url, ...options }) => {
+  fetch(url, options)
+    .then(processStatus)
+    .then(processBody)
+}
+
+const performRequest = (url, options) => {
+  const request = prepareRequest({ url, ...options })
+  return executeRequest(request)
```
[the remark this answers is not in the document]
viktor-yakubiv: I will keep it separate.

api/index.js
Joozty: No, not really. There was no reason I just did it because I wanted to have a unified error. Feel free to remove it. You can remove [this property](https://github.com/oacore/dashboard/blob/api/api/errors.js#L6) too. It's not used.

viktor-yakubiv: There is no need in removing the property.

However, there is a need in removing the `catch` clause because it can catch fetch initialisation/execution errors which may not be reported.

viktor-yakubiv:
By the way, I enabled auto-removing branches after merge.

api/index.js
Joozty: For me, all unused code can (should be) removed and that property is not used anywhere. 😄

api/index.js
```diff
@@ -1,89 +1,99 @@
...
+
+  return (/application\/([\w.-]\+)?json/g.test(type)
+    ? response.json()
+    : response.blob()
+  ).then(data => ({ data, type, headers }))
+}
+
+// TODO: Remove later
+// Old type of error processing
+// eslint-disable-next-line no-unused-vars
```
Joozty: Why didn't you remove it?

viktor-yakubiv: It would be good to remove it but I decided to do it in the next refactoring step. It's only my personal weakness to delete what just created. Sorry

PR #192, opened 2020-03-18
Rename deposit dates navigation item

viktor-yakubiv:
I found it's quite hard to rename id's and change the URL to reflect this change. We will have it deposit-dates everywhere.

### DIFF
--- a/texts/navigation.yml
+++ b/texts/navigation.yml
@@ -4,7 +4,7 @@ items:
   index: Overview
   data: Data
   content: Content
-  deposit-dates: Deposit dates
+  deposit-dates: Deposit compliance
   plugins: Plugins
   statistics: Statistics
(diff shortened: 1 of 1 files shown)


Joozty:
I would prefer to rename it everywhere. It shouldn't be so hard.

viktor-yakubiv:
It worth to take marketing decision immediately and then think about infrastructure.

Joozty:
Well, it takes max 10minutes to rename everything so why do it in new PR? I can do it. No problem.

viktor-yakubiv:
I do not feel expert in doing so. Also, I had bad experience doing a 30m task whole day yesterday.

If you want to rename, go for it. I do not think it's necessary for this moment and it may create breaking changes for other PRs.

PR #194, opened 2020-03-18
Table with Sidebar

Joozty:
@viktor-yakubiv This is the first part of moving to the new design of the table component. I tried to remove config and replace it with direct children. Also, I removed the bidirectional scrolling. It didn't work very well and I found it a potential issue for release. 

In the next part, I am going to develop the sidebar according to Figma design.


viktor-yakubiv:
If you see the issue for the release, can you please create separate stateless components for everything and test it on another page?

I did some basic testing and I am not really happy with it. I will express my experience a little bit later.

Joozty:
I don't see any issue for release. I just said I removed bidirectional scrolling because I wasn't happy with it. It was buggy and scrolling wasn't that smooth.

components/table/Sidebar.jsx
```diff
@@ -0,0 +1,48 @@
...
+    super(props)
+    const { children } = props
+    if (children) React.Children.only(children)
+  }
+
+  render() {
+    const { isOpen, children, ...passProps } = this.props
+    return (
+      <div className={tableClassNames.sidebar}>
+        {children && React.cloneElement(children, passProps)}
```
viktor-yakubiv: It would work only if the children is a single element. If it's more than one element, it will probably throw or pass props in bulk. You have no validation so I am not sure it's what you want to.

components/table/Sidebar.jsx
```diff
@@ -0,0 +1,48 @@
...
+      </div>
+    )
+  }
+}
+
+Sidebar.displayName = 'TableSidebar'
```
viktor-yakubiv: It's not TableSidebar at all. It's a separate component not related to table. Maybe something like _DetailsPane_

components/table/Sidebar.jsx
```diff
@@ -0,0 +1,48 @@
...
+    <div {...passProps}>
+      {children}
+      <CloseIcon
+        ref={closeIconRef}
+        aria-label="Close table sidebar"
+        onClick={() =>
+          closeIconRef.current.dispatchEvent(
+            new Event('sidebar-close', { bubbles: true })
+          )
+        }
```
viktor-yakubiv: Why do you create a custom event instead using standard click event?

package.json
```diff
@@ -52,6 +52,7 @@
     "d3-collection": "^1.0.7",
     "dayjs": "^1.8.21",
     "extract-css-chunks-webpack-plugin": "^4.7.4",
+    "hoist-non-react-statics": "^3.3.2",
```
viktor-yakubiv: Why do you need this?

pages/data-providers/[data-provider-id]/content.jsx
```diff
@@ -1,4 +1,5 @@
-import React, { useMemo } from 'react'
+// eslint-disable-next-line max-classes-per-file
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/content.jsx]: Prefer `/* eslint-disable max-classes-per-file */` in such cases. Why do you put it in the top?

package.json
Joozty: When a component is wrapped within HOC component like [this for example](https://github.com/oacore/dashboard/blob/master/utils/withErrorBoundary.jsx#L5). All static properties are lost.

Concretely I needed it for using `<Table.Sidebar />` because table component uses its own error boundary.

pages/data-providers/[data-provider-id]/content.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/content.jsx]: It's where my IDE puts it. Nonetheless it's gonna be removed in https://github.com/oacore/dashboard/pull/196

components/table/Column.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+class Column extends React.Component {
+  render() {
+    const { context, getter, id } = this.props
+    const value = getter ? getter(context) : context[id]
+    return value !== undefined ? value : null
```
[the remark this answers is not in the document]
Joozty: because it makes the rendering of table columns [much easier](https://github.com/oacore/dashboard/blob/table/pages/data-providers/%5Bdata-provider-id%5D/content.jsx#L100)...

components/table/Sidebar.jsx
Joozty: How so? Validation is in the constructor. I know if you change sidebar during runtime it will probably fail but for now, I consider as an edge case and will fix it when we actually face this problem.

components/table/Sidebar.jsx
Joozty: This line is just relic cod and it's gonna be removed in https://github.com/oacore/dashboard/pull/196. 

>  It's a separate component not related to the table. 

That's questionable. styling is tight up with a table component.

components/table/Sidebar.jsx
Joozty: How would I distinguish this event from row click? I would need to pass some ID or className and it's almost the same as this one. :)

Joozty:
> Second click on a row should not close the sidebar. It's not evident behaviour.
It doesn't close the sidebar. Are you sure about it? 

> I am not sure it's okay that try to scroll with cursor on the sidebar by trackpad/mouse scrolls the whole view. What do you think?

I like it (but I am one out of many). Anyway, this is something we should really decide on also it should be fixed in extra PR as we already agreed.

components/table/Sidebar.jsx
viktor-yakubiv: Sorry, I missed the validation in the sidebar. I did not read the code carefully enough.

components/table/Column.jsx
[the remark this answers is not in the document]
viktor-yakubiv: Yes, sorry for this comment. I found this later. I read files one by one during review what is not always good sequence.

package.json
viktor-yakubiv: I consider this being workaround and do not recommend to do so.

components/table/Sidebar.jsx
viktor-yakubiv: Yes, you may consider it questionable. This is the only place you see the component and it's tightly coupled to the Table. However, as a software engineer you are the designer of the application too. To my mind, if there is no reason to tie 2 components, why do it just because it looks so. Software design should not reflect UI design 1-by-1. UI design may change often but software should not change as often.

viktor-yakubiv:
> It doesn't close the sidebar. Are you sure about it?

Check the [screen recording](https://vimeo.com/upload#_=_).

> should be fixed in extra PR

👍

components/table/Sidebar.jsx
viktor-yakubiv: It's not correct that you render details window right in the table. The point of redesign was that it's not neither logical nor semantically correct to have not-a-row in the table.

package.json
Joozty: If you can find a better solution do it. This is my way how I fixed. That library is downloaded almost 9M times per week. I don't consider this as a workaround.

Joozty: It's even recommended by React docs
https://reactjs.org/docs/higher-order-components.html#static-methods-must-be-copied-over

components/table/Sidebar.jsx
Joozty: Sorry I was not clear. Sidebar is not a part of <table>. Table component consist of two divs. One contains table element and one contains sidebar.

Joozty:
Sorry, I should have read more carefully. I read double click instead of the second click. Fixed here https://github.com/oacore/dashboard/pull/194/commits/b53409905081f2f39e9a874f041c0e795ad32860

viktor-yakubiv:
@valerapro let us please know when the `/user` endpoint on development server works again. I am going to merge this PR but I cannot test it now.

valerapro:
Fixed. You can try

PR #196, opened 2020-03-19
Table sidebar styling

Joozty:


components/table/TableRow.jsx
```diff
@@ -1,15 +1,19 @@
 import React from 'react'
 import { Table } from '@oacore/design'
 import { classNames } from '@oacore/design/lib/utils'
+import { observer } from 'mobx-react'
 
 import tableClassNames from './index.css'
 
-const TableRow = React.memo(props => {
+const TableRow = observer(props => {
```
viktor-yakubiv: I am not sure you need `observer`. Does not it work without?

Joozty: Already tried it has to be there. Observer checks only direct children in render.  :) 

> Note that observer only subscribes to observables used during the own render of the component. So if observables are passed to child components, those have to be marked as observer as well. This also holds for any callback based components.

See docs for more
https://mobx.js.org/refguide/observer-component.html

viktor-yakubiv: So, there are 2 other options to consider: use decorator or wrap on export clause.

components/document-link/index.jsx
```diff
@@ -1,22 +1,21 @@
...
+
+import { Button, Icon } from 'design'
 
 const DocumentLink = React.memo(({ href, children }) => (
-  <a
+  <Button
+    tag="a"
     href={href}
-    // We want to pass referrer tag to our website
-    // eslint-disable-next-line react/jsx-no-target-blank
```
viktor-yakubiv: Oh, interesting issue. I see we have to add `Button` to the list of components that the rule has to check. This comment should stay.

components/table/Sidebar.jsx
```diff
@@ -43,6 +45,21 @@ const SidebarHeader = ({ children, ...passProps }) => {
   )
 }
 
+const SidebarBody = ({ children, ...passProps }) => (
```
viktor-yakubiv: This seems to be useless. I would prefer to avoid this and just inline children. Minor, I will probably fix it myself later.

design/detail-list/index.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import { classNames } from '@oacore/design/lib/utils'
+
+import styles from './styles.css'
+
+const DetailList = React.memo(({ children, className, ...passProps }) => (
```
viktor-yakubiv: `Details` and `Details.Item`? I am not sure

design/detail-list/index.jsx
```diff
@@ -0,0 +1,24 @@
...
+    {...passProps}
+  >
+    {children}
+  </ul>
+))
+
+const DetailItem = React.memo(({ name, className, children, ...passProps }) => (
+  <li className={classNames.use([styles.item]).join(className)} {...passProps}>
+    <div className={styles.header}>{name}</div>
+    <div className={styles.content}>{children}</div>
```
viktor-yakubiv: This may be extra element too. Do you apply any styling on it?
```suggestion
    {children}
```

store/works.js
```diff
@@ -10,17 +10,17 @@ class Works extends Pages {
...
     const [pageNumber, workId] = rowId.split('-', 2)
     // TODO: consider moving this endpoint to /data-providers/works/
-    await apiRequest(`/works/${workId}`, {
-      method: 'PATCH',
-      body: { disabled: !isVisible },
-    })
+    // await apiRequest(`/works/${workId}`, {
+    //   method: 'PATCH',
+    //   body: { disabled: !isVisible },
+    // })
```
viktor-yakubiv: How did this comment appear?

components/document-link/index.jsx
Joozty: is it possible to configure it?

components/table/Sidebar.jsx
Joozty: it is there just because it looks cool. 😄 (Not a good reason)

components/document-link/index.jsx
viktor-yakubiv: Check the [custom link components option](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md#custom-link-components).

Also, we may want to enable `allow-referrer` option.

store/works.js
Joozty: Once I finish I am gonna uncomment it. I am just afraid to call that endpoint while testing because I may lose track of which article I disabled. Hopefully, if we have a separate database for staging I won't need to do this while developing.

components/document-link/index.jsx
Joozty: That's quite clever 👍

design/detail-list/index.jsx
Joozty: Yes there is a styling applied for center vertically, i.e. icon with text match together.

design/detail-list/index.jsx
Joozty: Yep maybe. If you want you can rename it in your PR. :)

PR #197, opened 2020-03-19
Fix activity sidebar width

Joozty:


viktor-yakubiv:
It does not look good. I would think about it and probably organise another name changing. If so, this will get reverted.

PR #198, opened 2020-03-19
Read configuration from environment

viktor-yakubiv:
Introduces 3 configuration environments:
* production
* development
* local

where `local` environment is a `development` one targeted to local version of the API.

Adds a `dev-api` command to test API locally during development. @valerapro this may be useful for you.

Adds an identity provider URL as `IDP_URL`.

Replace `getLoginPage` with pure constant called `LOGIN_URL` and moves it to the central configuration file. The function logic moved to App component.

Moves `LOGOUT_URL` to the configuration file leaving `continue` parameter passing in the Logout component.

---

Importing configuration is **deprecated**. All configuration variables are passed via `process.env`.

---

Also, marks Now as a development area.

### DIFF
--- a/components/application/logout.jsx
+++ b/components/application/logout.jsx
@@ -1,28 +1,30 @@
 import React from 'react'
 import { AppBar } from '@oacore/design'
 
-import { API_URL } from '../../config'
 import styles from './logout.css'
 
 import { withGlobalStore } from 'store'
 
-const LOGOUT_URL = `${API_URL.replace(
-  '/internal',
-  ''
-)}/logout?continue=${encodeURIComponent(
-  typeof window !== 'undefined'
-    ? `${window.location.origin}/?reason=logout`
-    : ''
-)}`
+const { IDP_URL } = process.env
+
+const getLogoutUrl = pathname => {
+  const url = new URL('./logout', IDP_URL)
+  if (pathname != null) url.searchParams.set('continue', pathname)
+  return url
+}
 
 const Logout = ({ store }) => {
   if (!store.user) return null
   const name = store.user.email
+  const redirectUrl =
+    typeof window != 'undefined'
+      ? `${window.location.origin}?reson=logout`
+      : null
 
   return (
     <AppBar.Item className={styles.container}>
       <span className={styles.name}>{name}</span>{' '}
-      <a href={LOGOUT_URL}>Logout</a>
+      <a href={getLogoutUrl(redirectUrl)}>Logout</a>
     </AppBar.Item>
   )
 }
--- a/config.js
+++ b/config.js
@@ -1,18 +1,30 @@
-export const API_URL =
-  process.env.NODE_ENV === 'production'
-    ? 'https://api.core.ac.uk/internal'
-    : 'https://api.dev.core.ac.uk/internal'
+const NODE_ENV = process.env.NODE_ENV || 'development'
 
-export const getLoginPage = (path = window.location.origin) => {
-  if (process.env.NODE_ENV === 'production') {
-    return new URL(
-      `/login.html?continue=${encodeURIComponent(path)}`,
-      window.location.origin
-    ).toString()
-  }
-
-  return new URL(
-    `/test/login?continue=${encodeURIComponent(path)}`,
-    API_URL
-  ).toString()
+const local = {
+  API_URL: 'http://127.0.0.1:8000/internal',
+  IDP_URL: 'http://127.0.0.1:8000',
 }
+
+const development = {
+  API_URL: 'https://api.dev.core.ac.uk/internal',
+  IDP_URL: 'https://api.dev.core.ac.uk',
+}
+
+const production = {
+  API_URL: 'https://api.core.ac.uk/internal',
+  IDP_URL: 'https://api.core.ac.uk',
+}
+
+const validate = config =>
+  ['API_URL', 'IDP_URL'].forEach(param => {
+    if (config[param] == null) throw new Error(`${param} is not configured.`)
+  })
+
+const env = { local, development, production }
+const config = {
+  ...env.production,
+  ...env[NODE_ENV],
+}
+
+validate(config)
+module.exports = config
--- a/next.config.js
+++ b/next.config.js
@@ -7,7 +7,11 @@ const withSourceMaps = require('@zeit/next-source-maps')
 
 const camelCaseLoader = path.join(__dirname, 'webpack/camelcase-loader.js')
 
+const envConfig = require('./config')
+
 const nextConfig = {
+  env: envConfig,
+
   webpack(config, options) {
     const { dev, isServer } = options
 
--- a/pages/_app/index.jsx
+++ b/pages/_app/index.jsx
@@ -8,12 +8,13 @@ import '@oacore/design/lib/index.css'
 import Route from './route'
 import styles from './index.css'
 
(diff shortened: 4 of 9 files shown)


viktor-yakubiv:
I had a little bit more thinking about it and my solution is not nice. I want to move now.sh to development version of api and the rest on production version.

I found that running of development build of React is not a security issue but only [a performance issue](https://reactjs.org/docs/optimizing-performance.html#use-the-production-build). If we do not care about performance and bundle size on development builds, I propose to pass `NODE_ENV=development` to `now.json`. What do you think @Joozty?

Joozty:
I don't mind it. I am just worried because I don't run production on the localhost and now.sh is the only point where can I test if everything works fine. More precisely I am worried about API. Let's be honest here. We have had issues with API and if something works on `api.dev` it doesn't necessarily mean it works on prod API. It would make me sad if we discover the issue after release because production wasn't tested well in our review process. If at least dashboard.oacore.ac.uk could run production that would be great.

viktor-yakubiv:
I want to setup 3 environments as we discussed before:
- `dashboard-dev`evelopment on Now.sh
- `dashboard-pre`release and `dashboard-prod`uciton on AWS

viktor-yakubiv:
@mcancellieri and I just set the staging server up at [dashboard-pre.core.ac.uk](https://dashboard-pre.core.ac.uk) 🎉

config.js
```diff
@@ -1,18 +1,36 @@
...
-    return new URL(
-      `/login.html?continue=${encodeURIComponent(path)}`,
-      window.location.origin
-    ).toString()
-  }
+const local = {
+  API_URL: 'http://127.0.0.1:8000/internal',
+  IDP_URL: 'http://127.0.0.1:8000',
+  LOGIN_URL: 'http://127.0.0.1:8000/test/login',
+  LOGOUT_URL: 'http://127.0.0.1:8000/logout',
```
viktor-yakubiv: @Joozty are not `LOGIN_URL` and `LOGOUT_URL` obsolete?

Joozty: Yep they can be removed

viktor-yakubiv: Is it automatically discovered from IDP URL? Could you please point me where I can find how it's actually work?

components/application/logout.jsx
```diff
@@ -22,7 +20,7 @@ const Logout = ({ store }) => {
   return (
     <AppBar.Item className={styles.container}>
       <span className={styles.name}>{name}</span>{' '}
-      <a href={LOGOUT_URL}>Logout</a>
+      <a href={getLogoutUrl(window?.location.origin)}>Logout</a>
```
Joozty: You will need to pass `?reason=logout` to the URL otherwise, logout message doesn't appear.

config.js
Joozty: The only thing you should change is [here](https://github.com/oacore/dashboard/blob/master/pages/_app/index.jsx#L208) For the logout redirection param should look like `continue={window.location.origin}?reason=logout`.

components/application/logout.jsx
viktor-yakubiv: Thank you, I will fix this

config.js
viktor-yakubiv: It's not so according to [the implementation](https://github.com/oacore/dashboard/pull/198/files#diff-ac9542b946ce084e7a1b3976bd207257R84). It may be either my fault or artefact from #186.

Could you please double check it and help me to resolve this. What direction should I go?

Joozty: Simply just remove all changes you did in logout.jsx. And replace API_URL with IDF_URL. Obviously you can extract `LOGOUT_URL` into a function but the logic should stay the same.

Overall it should look like

```
import React from 'react'

import { IDP_URL } from 'config'
import styles from './logout.css'

const LOGOUT_URL = `${IDP_URL}/logout?continue=${encodeURIComponent(
  typeof window !== 'undefined'
    ? `${window.location.origin}/?reason=logout`
    : ''
)}`

const Logout = ({ store }) => {
  if (!store.user) return null
  const name = store.user.email

  return (
    <AppBar.Item className={styles.container}>
      <span className={styles.name}>{name}</span>{' '}
      <a href={LOGOUT_URL}>Logout</a>
    </AppBar.Item>
  )
}

export default withGlobalStore(Logout)
```

Does that make sense?

viktor-yakubiv: Thank you!

viktor-yakubiv:
@Joozty I cannot log in. What may be wrong?

components/application/logout.jsx
```diff
@@ -1,28 +1,28 @@
...
-  ''
-)}/logout?continue=${encodeURIComponent(
-  typeof window !== 'undefined'
-    ? `${window.location.origin}/?reason=logout`
-    : ''
-)}`
+const { IDP_URL } = process.env
+
+const getLogoutUrl = pathname => {
+  const url = new URL('./logout', IDP_URL)
```
Joozty: Out of curiosity. What's the point of having here dot? It looks like it has no effect at all.


```jsx
new URL('./logout', 'https://localhost:8000/test').toString() // "https://localhost:8000/logout"
```

next.config.js
```diff
@@ -7,7 +7,11 @@ const withSourceMaps = require('@zeit/next-source-maps')
 
 const camelCaseLoader = path.join(__dirname, 'webpack/camelcase-loader.js')
 
+const envConfig = require('./config')
+
 const nextConfig = {
+  env: envConfig,
```
Joozty: Would be nice to include [these variables](https://github.com/oacore/dashboard/blob/master/next.config.js#L127-L132) in config too. (We can do it in another PR if you want)

pages/_app/index.jsx
```diff
@@ -79,8 +80,10 @@ class App extends NextApp {
     }
   }
 
-  redirectToLogin = path => {
-    window.location.replace(getLoginPage(path))
+  redirectToLogin = pathname => {
+    const url = new URL('/')
```
Joozty: This will fail. You need to pass full URL or pass base as a second param.

```
new URL('/')
VM131:1 Uncaught TypeError: Failed to construct 'URL': Invalid URL
    at <anonymous>:1:1
```

pages/_app/index.jsx
```diff
@@ -79,8 +80,10 @@ class App extends NextApp {
...
+    url.searchParams.set('continue', pathname)
```
Joozty: Could you please add here `reason=logout` too?

components/application/logout.jsx
viktor-yakubiv: It would have effect with

```js
new URL('./logout', 'https://localhost:8000/test/').toString()
```

pages/_app/index.jsx
viktor-yakubiv: I guess it's server side rendering problem, is not it?

next.config.js
viktor-yakubiv: I will check if it's not passed implicitly. If it's not we should work more on this

viktor-yakubiv: I feel okay with current configuration. Do you mind to keep it?

Joozty: I think all config variables should be in one place for the sake of simplicity. I don't insist on doing it now. It can be done in separate PR and I can do it as the code is closer to me more.

viktor-yakubiv: Let's do to separately

PR #200, opened 2020-03-20
Fix disabled deposit dates export button

viktor-yakubiv:
Adds an ExportButton that renders either Anchor or Button based on props.


viktor-yakubiv:
It's not obvious behaviour for the pure component so I would not like to move it. Its basically workaround, the proper fix does not require element change

PR #203, opened 2020-03-24
Fix API request when method is HEAD

Joozty:


store/deposit-dates.js
```diff
@@ -62,7 +62,7 @@ class DepositDates {
   loadDepositDatesCount = async () => {
     try {
       this.isExportDisabled = false
-      const { headers } = await apiRequest(this.datesUrl, { mehtod: 'HEAD' })
```
viktor-yakubiv: Thank you for fixing this.

Joozty:
It fails without that condition. The body is not processed as a blob but as a json which is wrong. You sure you tested it correctly?

viktor-yakubiv:
I tested HEAD request to Google if I remember so I can be sure that Google does not append `Content-Type` to empty response. We should not do it too @valerapro. However, if it works so, we have no other way for now. Thank you for fixing this.

Joozty:
HEAD should mirror GET response except for the body, shouldn't it? Anyway, I am not expert in this. So whatever Google does is probably correct. 😄

PR #204, opened 2020-03-24
Fix sidebar not closing after navigation

viktor-yakubiv:
@Joozty I am not sure this solution is good. Do you have in mind anything better?

### DIFF
--- a/components/layout/side-bar.jsx
+++ b/components/layout/side-bar.jsx
@@ -12,15 +12,29 @@ const SideBar = ({
   className,
   tag: Tag = 'nav',
   id, // prevent passing ID to element attributes
+  onClick,
   ...restProps
 }) => {
   const [state, dispatch] = useContext(LayoutContext)
+
+  const close = (event) => {
+    const controlElementSelector = 'a, button, [type=submit], [type=button]'
+    const controlElement = event.target.closest(controlElementSelector)
+    if (controlElement != null) dispatch({ type: 'toggle_sidebar' })
+  }
+
+  const handleClick = (event) => {
+    if (onClick) onClick(event)
+    close(event)
+  }
+
   return (
     <Tag
       id={state.sidebarId}
       className={classNames
         .use(styles['side-bar'], state.sidebarOpen && styles.open)
         .join(className)}
+      onClick={handleClick}
       {...restProps}
     >
       <AppBar className={styles.sideBarHeader}>


Joozty:
Finally, somebody looked into this.

What about something like this. It also handles the case when the user clicks outside of the sidebar. (I haven't tested it properly but you probably see my point)


```jsx
import React, { useContext, useRef, useEffect } from 'react'

import LayoutContext from './context'
import styles from './styles.css'

import { Logo, CloseIcon } from 'design'

const SideBar = ({
  children,
  className,
  tag: Tag = 'nav',
  id, // prevent passing ID to element attributes
  onClick,
  ...restProps
}) => {
  const [state, dispatch] = useContext(LayoutContext)
  const sidebarRef = useRef(null)

  useEffect(() => {
    if (state.sidebarOpen) sidebarRef.current.focus()
    else sidebarRef.current.blur()
  }, [state.sidebarOpen])

  const handleClick = event => {
    if (onClick) onClick(event)
  }

  return (
    <Tag
      ref={sidebarRef}
      id={state.sidebarId}
      className={classNames
        .use(styles['side-bar'], state.sidebarOpen && styles.open)
        .join(className)}
      onClick={handleClick}
      onBlur={() => dispatch({ type: 'toggle_sidebar' })}
      tabIndex="0"
      {...restProps}
    >
      <AppBar className={styles.sideBarHeader}>
        <AppBar.Brand className={styles.sideBarHeaderBrand} href="/">
          <Logo />
        </AppBar.Brand>
        <AppBar.Item type="button" tag="button">
          <CloseIcon aria-label="Close sidebar" />
        </AppBar.Item>
      </AppBar>

      {children}
    </Tag>
  )
}

export default SideBar
```

viktor-yakubiv:
@Joozty the [`blur` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event) would not work, it's only for control elements such as inputs and buttons, also it does not bubble. The sidebar is a regular element which does not receive focus even if elements inside it do.

I would keep it as is and I will create an issue to handle click outside too.

Joozty:
That's why added there `tabIndex="-1"`. I tested it a little bit the other day and it worked just fine.

viktor-yakubiv:
@Joozty does not work for me, could you please check 506c11f for mistakes?

viktor-yakubiv:
@Joozty it was worth to try but I do not go this way and keep the solution as it was before with the following creating an issue about handling clicks outside. I share the solution with you, sou you can access the preview and test yourself:

- try to navigate using _Tab_ key; you will not be able to access individual links on the sidebar
- try to close sidebar using the close icon and then change the data provider; the sidebar unexpectedly opens

viktor-yakubiv:
I removed the commit but I experience the same behaviour as it was before 😕

viktor-yakubiv:
@Joozty can you please take this one?

viktor-yakubiv:
I retested this and do not see anymore the issues I faced before. I am merging this

Issue #188, opened 2020-03-16
Add a loading bar for identifying network requests


viktor-yakubiv:
I propose to add a global loading state reporting to identify network requests running. I created a [pen](https://codepen.io/yakubiv/pen/VwLxyaj) how this could look like.

The proposed component can be embedded into the _Application_ to display the loading. The Application component could accept `loading` prop that is boolean. Also, the `Progress` component can be rendered contextually, e.g. n the top of the _Table_.

@Joozty how could this be integrated into the store?

Joozty:
I can think of a very simple solution. Let me explain it. We will have loading property in the root store and this property will be simple increased/decreased by one. There is one caveat. The programmer is always responsible to clear loading state.

```jsx
class Root {
	isLoading = 0
	
	@action displayLoading() {
		this.isLoading += 1
	}

	@action hideLoading() {
		this.isLoading -= 1
	}
}
```

In the component, it would look like

```jsx
const Component = () => {
	useEffect(() => {
		try{
			store.setLoading()
			// api call here
		} finallly {
			store.hideLoading()
		}
	}, [])

	....
}

```

Loading component would look like:

```jsx
<LoadingBar loading={store.isLoading > 0} />
```

viktor-yakubiv:
Above the AppBar I guess.

I would design the store chunk slightly differently:

```js
class Root {
  requestsInProgress = 0
  
  @computed
  get loading() {
    return this.requestsInProgress > 0
  }
}
```

About the component implementation I do not agree. I thing that all the requests should be executed from the Store and the counter should be increased in the store itself.

Joozty:
I used actions because I read somewhere it's a good practice but to be honest I don't have any opinion about it. MobX allows [enforcing changing state only in actions](https://mobx.js.org/refguide/api.html#enforceactions). What do you think? 


Regarding the component implementation. I made it very general (for long-running tasks, not only API requests). If we want to have this only for API requests then I agree it should be only in the store.

PR #207, opened 2020-03-25
Add loading bar

Joozty:
I am not very satisfied with the solution but it was the best I could find. Alternatively, we could have some sort of before and after hooks for API request but this would result in duplicate code because it needed to be done for every API request separately.

@viktor-yakubiv If you have a better solution I am all ears.

Closes https://github.com/oacore/dashboard/issues/188

### DIFF
--- a/components/application/index.jsx
+++ b/components/application/index.jsx
@@ -1,10 +1,12 @@
 import React from 'react'
 
 import { Container, AppBar, SideBar, Main } from '../layout'
+import LoadingBar from '../loading-bar'
 import RepositorySelect from './repository-select'
 import ActivitySelect from './activity-select'
 import Head from './head'
 import Logout from './logout'
+import styles from './styles.css'
 
 import activities from 'store/activities'
 
@@ -13,11 +15,13 @@ const Application = ({
   dataProvider,
   activity,
   isAuthenticated = false,
+  isLoading = false,
   ...restProps
 }) => (
   <>
     <Head />
     <Container {...restProps}>
+      {isLoading && <LoadingBar className={styles.loadingBar} />}
       <AppBar variant={isAuthenticated ? 'internal' : 'public'}>
         {isAuthenticated ? (
           <>
--- a/store/root.js
+++ b/store/root.js
@@ -1,14 +1,28 @@
 import { observable, action, computed } from 'mobx'
 
 import activities from './activities'
+import Store from './store'
 import User from './user'
 import DepositDates from './deposit-dates'
 import Works from './works'
 
 import apiRequest from 'api'
 
-class Root {
-  @observable user = new User()
+class Root extends Store {
+  constructor() {
+    const request = async (...args) => {
+      this.requestsInProgress += 1
+      try {
+        return await apiRequest(...args)
+      } finally {
+        this.requestsInProgress -= 1
+      }
+    }
+
+    super(null, { request })
+  }
+
+  @observable user = new User(this.options)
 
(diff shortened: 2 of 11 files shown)


store/root.js
```diff
@@ -73,8 +89,8 @@ class Root {
     this.retrievePluginConfig()
 
     const url = `/data-providers/${this.dataProvider.id}`
-    this.works = new Works(url)
-    this.depositDates = new DepositDates(url)
+    this.works = new Works(url, this.apiRequest)
```
viktor-yakubiv: It's an options parameter, not a second initialisation parameter.

```suggestion
    this.works = new Works(url, this.options)
```

Despite of that, I propose to simply call it `request`

components/application/index.jsx
```diff
@@ -21,6 +22,7 @@ const Application = ({
       <AppBar isAuthorized={isAuthorized}>
         {dataProvider && <RepositorySelect value={dataProvider} />}
         <Logout />
+        <LoadingBar />
```
viktor-yakubiv: Even if the loading bar is on top of application bar, it has nothing to the application bar.

Joozty:
@viktor-yakubiv Go for it.

store/root.js
Joozty: Maybe [passing `this`](https://mobx.js.org/best/store.html#combining-multiple-stores) would be even better. Obviously the serialization would be harder but we don't need it now anyway.

components/application/index.jsx
Joozty: It's more about positioning and I liked it more having it below App bar.

viktor-yakubiv:
I like the loading bar more in the top. Maybe because it's more common. It's personal preference anyway. We can ask the team if you wish.

The goal is to make the only root component control everything and avoid side effects like `withGlobalStore` where it's possible.

To achieve the goal I had to make global `App` being `observer` @Joozty please let me know if you see significant performance issues.

If you think it's fine, feel free to rebase and merge.

Joozty:
> I like the loading bar more in the top. Maybe because it's more common. It's personal preference anyway. We can ask the team if you wish.

That's fine it looks even better now. 👍 

> The goal is to make the only root component control everything and avoid side effects like withGlobalStore where it's possible.

I just want to point out that using `withGlobalStore` and observer is basically the same. `withGlobalStore` adds just `store` prop on top of that but you know that already I guess. It just cannot be used on app component.


> To achieve the goal I had to make global App being observer @Joozty please let me know if you see significant performance issues.

I personally consider a very bad practice using `@observe` on root component. 

Let me explain. The application is rerendered 5 times more.
Every request triggers rerender - 4  request
One rerender is called when loading bar needs to be hidden

Like it's not a big deal because react is really good with virtual DOM but I still think only Loading bar should be observer.

And you removed my best feature - letting animation to finish before hiding it. Why? I was thinking it was quite cool. 😄

viktor-yakubiv:
Yes, this is what I really sorry for. On the other hand, I am not sure it's correct if animation goes when the data is fetched.

Joozty:
@viktor-yakubiv Sometimes I do things that are not even necessary 😄

Joozty:
What's the conclusion of this one? Are you going to change it or you want to leave it as is?

viktor-yakubiv:
I would leave this as is but I needed to do leftover here.

viktor-yakubiv:
I checked the code, I would leave this as is unless we have performance issues

PR #209, opened 2020-03-26
Fix postInstall script

Joozty:
Previously cp command created assets subdirectory in ./public/design if design folder existed. So before copying the folder is firstly removed. We could have used -T flag but it's not available on MacOS therefore this approach.

See more:
https://stackoverflow.com/questions/23698183/how-to-force-cp-to-overwrite-directory-instead-of-creating-another-one-inside

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,6 @@
     "lint:js": "eslint --ignore-path .gitignore --ext .js,.jsx .",
     "lint:css": "stylelint --ignore-path .gitignore '**/*.css'",
     "lint": "run-p lint:*",
-    "postinstall": "cp -r \"$(npm explore @oacore/design -- pwd)/lib/assets\" ./public/design"
+    "postinstall": "rm -rf ./public/design/ && cp -r \"$(npm explore @oacore/design -- pwd)/lib/assets\" ./public/design"
   }
 }


package.json
```diff
@@ -85,6 +85,6 @@
     "lint:js": "eslint --ignore-path .gitignore --ext .js,.jsx .",
     "lint:css": "stylelint --ignore-path .gitignore '**/*.css'",
     "lint": "run-p lint:*",
-    "postinstall": "cp -r \"$(npm explore @oacore/design -- pwd)/lib/assets\" ./public/design"
+    "postinstall": "rm -rf ./public/design/ && cp -r \"$(npm explore @oacore/design -- pwd)/lib/assets\" ./public/design"
```
viktor-yakubiv: I am not an expert here but I am not sure about `&&`. Chaining using `&&` fails if the first command fails (due to access error or any other one). If you still want the second one to make an attempt, prefer `;`.

I am not sure it would help because the only error I see is an access error which if happens, does it both commands.

```suggestion
    "postinstall": "rm -rf ./public/design/; cp -r \"$(npm explore @oacore/design -- pwd)/lib/assets\" ./public/design"
```

Joozty: I think -f covers that case. If `rm` fails it doesn't make sense to run `cp` command because it's likely that it will fail too or it will copy it incorrectly.

PR #211, opened 2020-03-26
Redirect users to the legacy Dashboard if access forbidden

viktor-yakubiv:
@oacore/dashboard it's poorly tested faking a login error. It **requires setting up CORS** for the `dashboard.core.ac.uk/login_check` endpoint in the same way we have on API to work properly.

## Algorithm

At this moment the feature works in the following way:

1. If the `api.core.ac.uk/user` returns 2xx, logged it
2. If the `api.core.ac.uk/login_check` returns 2xx, logged in
3. Else send the same data to `dashboard.core.ac.uk/login_check`
4. Without checking any response, does not matter was it successful or was there a CORS error, redirect the user to `dashboard.core.ac.uk`

## Last step explained

1. If the legacy dashboard logged user in using credentials we sent, the user after visiting the home page will be redirected to the internal area
2. If the user is not logged in because of CORS they are obliged to log in again

## Assumptions

I assume that the API returns 402/403 **only** when the user passed valid credentials but is not logged in only because of the billing plan. This means the user can easily login with the same credentials to the legacy Dashboard so I just redirect the same request to another endpoint. 

Also, I assume that the user is not logged into the API. If they are, refreshing the page or visiting the new Dashboard in the new tab will provide access to the new Dashboard.

@valerapro please confirm this works as I described.

### DIFF
--- a/public/login.js
+++ b/public/login.js
@@ -1,3 +1,22 @@
+/* eslint-disable max-classes-per-file */
+
+class UnauthorisedError extends Error {}
+class ForbiddenError extends Error {}
+
+function performLoginRequest(url, data) {
+  return fetch(url, {
+    method: 'POST',
+    body: data,
+    credentials: 'include',
+  }).then(response => {
+    if (response.status === 401)
+      throw new UnauthorisedError('Wrong credentials')
+    if (response.status === 402 || response.status === 403)
+      throw new ForbiddenError('User does not have access to this version')
+    return response
+  })
+}
+
 function addLoadingAnimation() {
   const form = document.getElementById('login-form')
   form.classList.add('loading')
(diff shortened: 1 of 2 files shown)


public/login.js
```diff
@@ -31,23 +50,33 @@ function login(event) {
   const urlParams = new URLSearchParams(window.location.search)
   const identityProviderUrl =
     urlParams.get('identity_provider_url') || 'https://api.core.ac.uk/'
+  const fallbackIDP =
+    urlParams.get('fallback_idp') || 'https://dashboard.core.ac.uk/'
```
Joozty: I guess this url param is never set so why to check it?

```
  const fallbackIDP = 'https://dashboard.core.ac.uk/'
```

public/login.js
```diff
@@ -31,23 +50,33 @@ function login(event) {
...
-    })
-    .finally(() => {
+  const loginUrl = new URL('login_check', identityProviderUrl)
+  performLoginRequest(loginUrl, data)
+    .then(
+      () => 'login-processing',
+      error => {
+        if (error instanceof ForbiddenError) {
+          const fallbackUrl = new URL('login_check', fallbackIDP)
+          return performLoginRequest(fallbackUrl, data).then(
```
Joozty:
```suggestion
          return performLoginRequest(fallbackUrl, data).finally(() => 'login-fallback')
```

public/login.js
viktor-yakubiv: I forget to set it but it works fine with the default value 🙂

public/login.js
viktor-yakubiv: What is the difference? I am not sure I can chain `then` to `finaly` and get the result from `finaly`'s callback.

Joozty: Never mind I misread the code. Sorry about that

mcancellieri:
I'll have a look (it's not so easy to replicate the same configuration because of different versions and different libraries used...) but I am thinking that a polite message and a link are less unexpected to the user that a magical login somewhere else?

viktor-yakubiv:
We need to add the following into the configuration file for the `dashboard.core.ac.uk`:

```apache
<LocationMatch "^/login_check\b">
  <IfModule mod_headers.c>
    SetEnvIf Origin "https?://([\w-]+\.)?(core.ac.uk|oacore.now.sh)$" AccessControlAllowOrigin=$0
    Header set Access-Control-Allow-Origin %{AccessControlAllowOrigin}e env=AccessControlAllowOrigin
    Header set Access-Control-Allow-Methods POST
    Header set Access-Control-Allow-Headers Content-Type
  </IfModule>
</LocationMatch>
```

I will leave an issue for this.

mcancellieri:
Can you try now?

viktor-yakubiv:
I pushed a testing commit. It does not work. We can proceed without CORS. You can check the behaviour on the preview link using any user.

mcancellieri:
This behaviour is ok for now... What else is needed to release this feature?

viktor-yakubiv:
I can drop the last commit and merge.

PR #214, opened 2020-03-27
Measure speed of endpoints

Joozty:
Maybe it would be better to have this measure on the backend but this is also good. This can help us understand the bottlenecks and proactively fix them instead of waiting for the user's feedback.

You can find measurement results [here](https://analytics.google.com/analytics/web/#/report/content-site-speed-user-timings/a66779096w222136427p211062458/_u.date00=20200326&_u.date01=20200327&userTimingsExplorer-segmentExplorer.segmentId=analytics.userTimingVariable&userTimingsExplorer-table.plotKeys=%5B%5D/).

### DIFF
--- a/pages/_app/index.jsx
+++ b/pages/_app/index.jsx
@@ -10,7 +10,7 @@ import Route from './route'
 import styles from './index.css'
 
 import { UnauthorizedError } from 'api/errors'
-import logPageView from 'utils/analytics'
+import { logPageView } from 'utils/analytics'
 import { initStore, GlobalProvider } from 'store'
 import Application from 'components/application'
 import { Sentry } from 'utils/sentry'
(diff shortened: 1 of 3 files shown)


api/index.js
```diff
@@ -79,7 +79,7 @@ const executeRequest = ({ url, ...options }) =>
     .then(processStatus)
     .then(response => processBody(response, { ...options }))
 
-const performRequest = (url, options) => {
+const performRequest = async (url, options) => {
```
viktor-yakubiv: No need in that. `async` basically converts function to Promise. If you return Promise, there is no need to call it async.

Joozty:
I moved it to the store level. I don't think passing request as an option to store since decorator function using store values. If you insist I can move it one level up.

Joozty:
The idea behind utils is to share it across modules. So I don't see anything wrong with it. The analytics functions should be mocked in tests anyway. 🙂

viktor-yakubiv:
It should not exist in test. Mock is needed only to check the behaviour but not to hide bad practice. Utils should not exist either ideally, technically it's the place where code goes that does not have a better place.

PR #218, opened 2020-03-30
Upgrade next

Joozty:
In the first commit, I removed the deprecated getInitialProps method and did appropriate changes. It was pretty straightforward. Nonetheless, the second commit is more interesting.

I decided to switch to the new postcss default support in NextJS. There were some breaking changes. I had to rename all `css` files to `module.css` and move variables to classes. `:root` selector is not supported. There is a [bug reported](https://github.com/zeit/next.js/issues/10142) so when it gets fixed we can move them back. (Maybe it is even better as is now. I find it as a correct solution). 

The reason why I decided to switch it was mainly because hot reaload didn't work very well and also we had some problems with css from design repo. This has been all fixed now.

We will need to postpone this. CSS in dev mode is imported incorrectly. See https://github.com/zeit/next.js/issues/10148


postcss.config.js
```diff
@@ -1,15 +1,9 @@
...
-  plugins: [
-    presetEnv({
+  plugins: {
+    'postcss-preset-env': {
       stage: 0,
-    }),
-    extend(),
-    nano({
+    },
+    'cssnano': {
```
viktor-yakubiv: If you configure it so, there is no need to use `.js`. You can make this file `.yml`, `.json` or `rc`

design/card/index.module.css
```diff
@@ -1,12 +1,10 @@
-:root {
+.container {
```
viktor-yakubiv: This is breaking change, I am not sure we can live with this. If you use `--card-padding-x`, for example, somewhere outside the card container, you do not get any value.

If you want to get rid you from `:root` consider the pseudoselector being `html` in our case. It's called root to avoid writing a list like `html, svg`, Please, replace `:root` with `html` with a TODO and the link to the bug.

next.config.js
```diff
@@ -11,83 +10,41 @@ const envConfig = require('./config')
...
-      test: /\.css$/,
-      include: /\/design\/lib/,
-      use: [
-        hotCss,
-        { ...cssLoader, options: { ...cssLoader.options, modules: false } },
-        postcssLoader,
-      ].filter(Boolean),
+      if (cssLoaderConfig && cssLoaderConfig.options) {
+        // Patch it with the new config
+        cssLoaderConfig.options.localsConvention = 'camelCase'
```
viktor-yakubiv: Why could not you pass an option to Next's `withCss`?

pages/_app/index.module.css
```diff
@@ -1,9 +1,3 @@
-html,
-body,
-:global(#__next) {
-  height: 100%;
-}
-
```
viktor-yakubiv: Where did this go? I cannot find it anywhere?

Joozty: https://github.com/oacore/dashboard/blob/upgrade-next/pages/_app/global.css

next.config.js
Joozty: Because `withCss` requires plugin. This solution doesn't require any additional packages. It's default one from [Next.js](https://nextjs.org/docs/advanced-features/customizing-postcss-config)

design/card/index.module.css
Joozty: > This is breaking change, I am not sure we can live with this. If you use --card-padding-x, for example, somewhere outside the card container, you do not get any value

Isn't it right> I mean that variable should be available only within Card scope or moved to global.css file.

I am afraid [none of pure selectors is allowed](https://github.com/zeit/next.js/issues/10142). We just need to wait till they fix it.

postcss.config.js
Joozty: I think nextjs supports only [.js or .json](https://nextjs.org/docs/advanced-features/customizing-postcss-config) which is probably the same.

Joozty:
> Why are you obliged to rename all the CSS file with .module suffix?

It's how nextjs distinguish postcss files from regular ones. I don't like it but it is what it is.

> How the design library is supported now?

It's guaranteed that design css files will be imported before any local ones.

design/card/index.module.css
viktor-yakubiv: It is right, it's how CSS works and how the `design` was designed. I do not want to have a central configuration file for it but it's considerable. I quite like [HiQ](https://jonathanharrell.github.io/hiq/)'s approach.

~If it's not possible, can you duplicate selector with linter disabling and a TODO?~ The current solution is fine then.

Joozty: Well, we need to wait anyway because local hot reloading reorders css files. So maybe it gets fixed along with https://github.com/zeit/next.js/issues/10148. If that's the case I will revert back to `:root` usage.

Sebastp:
any updates?

Joozty:
@Sebastp Not yet. We are still waiting for fixing this bug https://github.com/zeit/next.js/issues/11946

Joozty:
@viktor-yakubiv So the bug regarding CSS order has been just fixed. We can use the canary version and merge or wait for version 9.3.6. What do you prefer?

viktor-yakubiv:
If waiting drains you because of unclosed task, let's merge. Otherwise let's wait.

viktor-yakubiv:
🎉

PR #220, opened 2020-03-31
Remove @observer from the global App component

viktor-yakubiv:
Fixes (should fix) the bug with infinite redirect loop in 'back' button press.

Creates an observer LoadingBar component inside the Application component.


viktor-yakubiv:
It was a draft

PR #221, opened 2020-03-31
Remove @observer from the global App component

viktor-yakubiv:
Fixes (should fix) the bug with infinite redirect loop in 'back' button press.

Inside the Application component creates LoadingBar component connected to the store.


viktor-yakubiv:
Seems like it does the job.

PR #222, opened 2020-03-31
Jenkins pipelines

Joozty:


viktor-yakubiv:
I probably have an easier solution for this: subscribe for the repository updates in the channel

PR #224, opened 2020-03-31
Reset table size when search/order changes

Joozty:
fix for https://kmi-ou.atlassian.net/browse/CORE-3125

### DIFF
--- a/components/table/index.jsx
+++ b/components/table/index.jsx
@@ -173,7 +173,7 @@ class InfiniteTable extends React.PureComponent {
 
     this.setState(newState)
 
-    const data = await pages.slice(0, size)
+    const data = await pages.slice(0, newSize)
 
     this.setState({
       data,


components/table/index.jsx
```diff
@@ -173,7 +173,7 @@ class InfiniteTable extends React.PureComponent {
 
     this.setState(newState)
 
-    const data = await pages.slice(0, size)
+    const data = await pages.slice(0, newSize)
```
viktor-yakubiv: Where do you get `newSize`?

Joozty: How is it not clear? For example if the value of newSize is 1000 then `pages` instance will check how many records. If less<1000 it loads next page and checks if >=1000. If not it load another batch and checks again.

So, in this case, the newSize is reset to 100 records, i.e. only one page is loaded.

viktor-yakubiv:
I still have no idea how this is fixed but I trust your expertise.

Joozty:
I am afraid you will need to study table component more. It makes sense to me because I developed it but I understand that it may be hard to get it. 

Maybe I try one more time. When a user changes order or change search then table component generates `force` flag. That flag [removes all records stored in MobX](https://github.com/oacore/dashboard/blob/fix-table-reseting/store/helpers/pages/pages.js#L33) and requests new ones [via slice method](https://github.com/oacore/dashboard/blob/fix-table-reseting/store/helpers/pages/pages.js#L25). Since there was a bug I needed to change size variable. If `force` flag is generated then `newSize` is set to 100 (one page).

PR #253, opened 2020-04-01
Styling adjustments

Joozty:


pages/data-providers/[data-provider-id]/overview.jsx
```diff
@@ -12,18 +12,41 @@ import PerformanceChart from 'components/performance-chart'
 // eslint-disable-next-line
 import styles from './overview.css'
 
+const OverviewCard = ({ children, className, ...passProps }) => (
+  <Card
+    className={classNames.use([styles.overviewCard], className)}
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/overview.jsx:17]: Why do you use an array here?

components/performance-chart/styles.css
```diff
@@ -1,5 +1,6 @@
 .container {
   position: relative;
+  max-height: 300px;
```
viktor-yakubiv, line 3: Does not a single limit placement fix the issue?

components/performance-chart/styles.css
```diff
@@ -22,7 +23,6 @@
   position: absolute;
   top: 50%;
   left: 50%;
-  margin: 12.5% 0 0;
```
viktor-yakubiv, line 25: It's calculated value from dropped 25% in the parent container. Why do you remove this?

pages/data-providers/[data-provider-id]/overview.jsx
```diff
@@ -12,18 +12,41 @@ import PerformanceChart from 'components/performance-chart'
...
+const OverviewCard = ({ children, className, ...passProps }) => (
+  <Card
+    className={classNames.use(styles.overviewCard, className)}
+    {...passProps}
+  >
+    {children}
+  </Card>
+)
+
+const BrowseButton = ({
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/overview.jsx:24]: It's not `BrowseButton` is it? It's more `LinkButton`, `AnchorButton` or `ButtonLink`, all you do here is changing `tag` to `a`. However, you do not even require `href`

pages/data-providers/[data-provider-id]/overview.jsx
```diff
@@ -68,19 +89,19 @@ const DepositingCard = ({ chartData, complianceLevel }) => {
             amount: (100 - complianceLevel).toFixed(1),
           })}
         </p>
-        <Button variant="contained" href="deposit-dates" tag="a" disabled>
+        <BrowseButton href="deposit-dates" tag="a" disabled>
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/overview.jsx:92]: 1. `disabled` does not work with `a` yet
2. `BrowseButton` passes `tag=a`

pages/data-providers/[data-provider-id]/overview.jsx
```diff
@@ -143,7 +162,7 @@ const DashboardView = ({
     <DOICard
       outputsCount={metadataCount}
       doiCount={doiCount}
-      enrichmentSize={doiEnricmentSize}
+      enrichmentSize={doiEnrichmentSize}
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/overview.jsx:165]: Thank you for this fix.

pages/data-providers/[data-provider-id]/overview.jsx
```diff
@@ -39,7 +62,7 @@ const DataStatisticsCard = ({
   className,
   ...restProps
 }) => (
-  <Card className={className} {...restProps}>
+  <OverviewCard className={className} {...restProps}>
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/overview.jsx:65]: If you do not append own classes, `className` can be passed implicitly via `restProps`

pages/data-providers/[data-provider-id]/overview.css
```diff
@@ -19,3 +19,13 @@
...
   }
 }
+
+.overview-card {
+  display: flex;
+  flex-direction: column;
+}
+
+.browse-button {
+  align-self: flex-start;
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/overview.css:29]: You might wanted to say `flex-end`. However, it's not needed anyway when you use `margin-top: auto`. You need only one of these. I prefer `align-self: flex-end`.

pages/data-providers/[data-provider-id]/overview.css
```diff
@@ -19,3 +19,13 @@
     grid-template-columns: repeat(3, 1fr);
   }
 }
+
+.overview-card {
+  display: flex;
+  flex-direction: column;
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/overview.css:25]: You should combine this with `justify-content: flex-start`

pages/data-providers/[data-provider-id]/doi.jsx
```diff
@@ -107,7 +107,7 @@ const DepositDates = ({
         searchable
       >
         <Table.Column
-          id="repo_doi"
+          id="repoDoi"
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/doi.jsx:110]: Is `id` used in any way?
- If is, why?
- If it's not, can it be just `doi` or `origin-doi`?

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/doi.jsx:110]: It's used for sorting. The value is appended after sortBy parameter. Also it's used as a default getter for cell value.

components/performance-chart/styles.css
Joozty: Unfortunately not :(

pages/data-providers/[data-provider-id]/overview.css
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/overview.css:25]: I wanted to keep all other elements with width 100%.

pages/data-providers/[data-provider-id]/overview.css
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/overview.css:29]: `flex-end` would move the button to the right. It defines alignment along with cross not main axis.

pages/data-providers/[data-provider-id]/overview.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/overview.jsx:24]: I took the name from the text which this button always contains. But I will change it to LinkButoon.

components/performance-chart/styles.css
Joozty: It looked good on small screen but when screen gets bigger it started to move down to the chart. So I rather removed it.

pages/data-providers/[data-provider-id]/overview.css
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/overview.css:25]: It's align-items stretch in this case. This is default too.

Issue #223, opened 2020-03-31
Remove LinkedList


viktor-yakubiv:
`LinkedList` is significant overhead meaningfully and from the performance view.

Built-in `Array` has the same functionality so it can be easily achieved by wrapping or extending the `Array`. The interface can stay the same but the class should be renamed more appropriate like `ChunkArray` or `ChunkList`.

Joozty:
Can you please share with me the performance insights?

viktor-yakubiv:
I consider the current implementation being slow. I had just a feeling but your question sounded so picky I finally decided to take this challenge. And the **results made me wonder a lot**, the difference is much bigger than I expected.

### Algorithm

I used the benchmarking code with the following configuration:

- Bucket size: 100
- Count of buckets: 1000
- Count of operations: 1 000 000

The snippet itself:

```js
const CHUNK_SIZE = 1e2
const CHUNKS_COUNT = 1e3
const TEST_ITERATIONS = 1e6
const chunk = Array.from(Array(CHUNK_SIZE), (_, i) => i)

const testOne = (List, size = CHUNKS_COUNT, limit = TEST_ITERATIONS) => {

  const l = new List()
  for (let i = 0; i < size; ++i) l.add(chunk)
  
  const start = Date.now()
  let total = 0
  for (let i = 0; i < limit; ++i) {
    const index = (CHUNKS_COUNT - 1) * CHUNK_SIZE + 1
    let x = l.get(index)
    total = (total + x) % 100 // avoid optimisations
  }
  const end = Date.now()
  
  return [start, end]
}

const test = () => {
  const [ta1, ta2] = testOne(LinkedList)
  const [tb1, tb2] = testOne(ChunkList)
  const [tc1, tc2] = testOne(ExtendedArray)

  const ta = ta2 - ta1
  const tb = tb2 - tb1
  const tc = tc2 - tc1

  return {
    'LinkedList': ta, 
    'ChunkedList': tb,
    'Array': tc,
    'LinkedList / ChunkedList': ta / tb,
    'LinkedList / Array': ta / tc,
  }
}
```

### What's been benchmarked?

The `LinkedList` is used from the project.

The `ChunkedList` is a simple optimisation based on the built-in Array written by myself:

```js
class ChunkList {
  constructor(size) {
    this.chunkSize = size
    this.chunks = []
    this.done = false
  }

  add(chunk) {
    if (this.done) throw new Error('Only one chunk and have different size')
    this.chunks.push(chunk)
[...]

PR #254, opened 2020-04-02
Remove unused code

Joozty:
Closes https://github.com/oacore/dashboard/issues/223


viktor-yakubiv:
Do not you want to simplify Pages too? You had a good roadmap in #223.

Joozty:
I simplified it as much as I could. There is always something to improve but I consider it done for this moment. Maybe next time. 🙂

PR #256, opened 2020-04-07
Chart.js

viktor-yakubiv:
@Joozty I tried to replace TimeLagChart with Chart.js. It's not as easy as I expected. Actually, it's hard to properly configure Chart.js too. However, I have to admit, it works faster.

My goals are:
- [x] performance
- [ ] style configuration from CSS
- [ ] simple configuration for any chart right from props
- [x] accessibility (possible via `children`)

For the second and third I can develop some kind of API. You may like it, may not. I appreciate any feedback. However, the biggest problem I see 🚨 I recreate the Chart instance at any time when props (references) change.

Meanwhile, can you please take a look at the current output?

### DIFF
--- a/charts/bar-chart.jsx
+++ b/charts/bar-chart.jsx
@@ -1,0 +1,54 @@
+import React from 'react'
+
+import Chart from './chart'
+
+const BarChart = ({ data, children, options, ...chartProps }) => {
+  const labels = data.map(([x]) => x)
+  const values = data.map(([, y]) => y)
+
+  // eslint-disable-next-line
+  const barData = {
+    labels,
+    datasets: [
+      {
+        label: 'Deposit time lag',
+        data: values,
+        backgroundColor: '#8bc34a',
+      },
+    ],
+  }
+
+  const chartOptions = {
+    scales: {
+      xAxes: [
+        {
+          // type: 'linear',
+          gridLines: {
+            // display: false,
+          },
+          ticks: {
+            stepSize: 1,
+          },
+        },
+      ],
+      yAxes: [
+        {
+          type: 'linear',
+          display: false,
+          gridLines: {
+            display: false,
+          },
+        },
+      ],
+    },
+    ...options,
+  }
+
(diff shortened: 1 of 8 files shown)


viktor-yakubiv:
@Joozty I played a bit with my code. If you check the [`charts/styles.css`](https://github.com/oacore/dashboard/pull/256/files#diff-0f9e528ee46203868fcd9f5a82d766a7) this is my way to configure the colours of the chart.

It's not pretty, I do not like it really, just wanted to try. It works on initial render and does not listen to CSS changes. Unfortunately, there is no wat to listen to it from `getComputedStyle()` result. However, I am sure there should be a way to listen to canvas node CSS changes if we want it.

viktor-yakubiv:
I explored [existing React wrapper](https://github.com/jerairrest/react-chartjs-2/blob/73a47d1d748df986ba8f96f4afe3cc7d4d09a7e3/src/index.js#L240), they do not do it differently. Basically, the author destroys the chart instance if something changes and creates a new one. This is exactly what I do.

Joozty:
I wouldn't say it's bad. It's just a different way of doing it in comparison with recharts. I must admit that the speed is quite notable. What happens when props change? Is there any flickering or is it smooth? DO you think you can manage to style the charts the same way as are now? I mean would there be any difference in look if we switched to it?

viktor-yakubiv:
The might be slight changes in the bar chart. Also, I think that I will replace the doughnut chart and the horizontal bar chart for DOIs. Of course, I will do my best to improve usability.

In terms of code, based on complex Chart.js configuration I see no other way to configure charts than creating children elements what I do not personally like but can stick with it.

```jsx
<Chart type="bar" labels={['Jan', 'Feb', 'Mar']}>
  <Dataset label="Income" data={[1.2, 1.3, 1.2]} />
</Chart>
```

viktor-yakubiv:
It's worth to close this. I hope we took enough output from this

PR #258, opened 2020-04-08
Change password

Joozty:
This PR adds change password form in settings tab. Password change is handled in an iframe.

### DIFF
--- a/components/forms/change-password/index.jsx
+++ b/components/forms/change-password/index.jsx
@@ -1,0 +1,43 @@
+import React, { useRef, useCallback, useEffect } from 'react'
+
+import styles from './styles.css'
+
+import { Card } from 'design'
+import { withGlobalStore } from 'store'
+
+const { IDP_URL } = process.env
+
+const ChangePassword = ({ className, store, tag }) => {
+  const iframeRef = useRef(null)
+  const email = store.user?.email
+  const searchParams = {
+    ...(email != null ? { email } : {}),
+    identity_provider_url: IDP_URL,
+  }
+
+  const search = new URLSearchParams(searchParams).toString()
+  const resize = useCallback(() => {
+    iframeRef.current.style.height = `${iframeRef.current.contentWindow.document.body.offsetHeight}px`
+  })
+
+  useEffect(() => {
+    const observer = new ResizeObserver(resize)
+    observer.observe(iframeRef.current.contentWindow.document.body)
+    return () => observer.disconnect()
+  }, [])
+
+  return (
+    <Card className={className} tag={tag}>
+      <h2>Change password</h2>
+      <iframe
+        ref={iframeRef}
+        className={styles.changePasswordIframe}
+        title="Password change form"
+        src={`/secure/reset.html?${search}`}
+        onLoad={resize}
+      />
+    </Card>
+  )
+}
+
+export default withGlobalStore(ChangePassword)
--- a/components/forms/change-password/index.jsx
+++ b/components/forms/change-password/index.jsx
@@ -1,14 +1,14 @@
-import React, { useRef, useCallback, useEffect } from 'react'
+import React from 'react'
 
 import styles from './styles.css'
 
+import { IframeForm } from 'components/forms'
 import { Card } from 'design'
 import { withGlobalStore } from 'store'
 
 const { IDP_URL } = process.env
 
 const ChangePassword = ({ className, store, tag }) => {
-  const iframeRef = useRef(null)
   const email = store.user?.email
   const searchParams = {
     ...(email != null ? { email } : {}),
@@ -16,25 +16,14 @@ const ChangePassword = ({ className, store, tag }) => {
   }
 
   const search = new URLSearchParams(searchParams).toString()
-  const resize = useCallback(() => {
-    iframeRef.current.style.height = `${iframeRef.current.contentWindow.document.body.offsetHeight}px`
-  })
-
-  useEffect(() => {
-    const observer = new ResizeObserver(resize)
-    observer.observe(iframeRef.current.contentWindow.document.body)
-    return () => observer.disconnect()
-  }, [])
 
   return (
     <Card className={className} tag={tag}>
       <h2>Change password</h2>
-      <iframe
-        ref={iframeRef}
+      <IframeForm
         className={styles.changePasswordIframe}
         title="Password change form"
         src={`/secure/reset.html?${search}`}
-        onLoad={resize}
       />
     </Card>
   )
--- a/components/forms/change-password/styles.css
+++ b/components/forms/change-password/styles.css
@@ -1,0 +1,4 @@
+.change-password-iframe {
+  width: 100%;
+  max-width: 50rem;
+}
--- a/components/forms/change-password/styles.css
+++ b/components/forms/change-password/styles.css
@@ -1,4 +1,8 @@
 .change-password-iframe {
+  --card-padding-x: 0;
+  --card-padding-y: 0;
+  --card-box-shadow: 0;
+
   width: 100%;
   max-width: 50rem;
 }
--- a/components/forms/index.js
+++ b/components/forms/index.js
@@ -1,1 +1,2 @@
 export ChangePassword from './change-password'
+export IframeForm from './iframe-form'
--- a/components/forms/index.js
+++ b/components/forms/index.js
@@ -1,0 +1,1 @@
+export ChangePassword from './change-password'
--- a/pages/_app/index.jsx
+++ b/pages/_app/index.jsx
@@ -32,7 +32,7 @@ class App extends NextApp {
     logPageView(url)
     const { dataProvider, activity } = new Route(url)
 
-    if (dataProvider) {
+    if (!Number.isNaN(Number(dataProvider))) {
       this.store.switchDataProvider(dataProvider)
       this.store.changeActivity(activity || 'overview')
     }
--- a/pages/_app/index.jsx
+++ b/pages/_app/index.jsx
@@ -28,6 +28,8 @@ class App extends NextApp {
     isAuthorized: false,
   }
 
+  store = initStore()
+
   handleRouteChange = (url) => {
     logPageView(url)
     const { dataProvider, activity } = new Route(url)
@@ -114,15 +116,11 @@ class App extends NextApp {
   }
 
   async componentDidMount() {
-    const store = initStore()
     const { router } = this.props
 
     // we use routeChangeStart to immediately update selected activity
     router.events.on('routeChangeStart', this.handleRouteChange)
     window.addEventListener('unhandledrejection', this.handlePromiseRejection)
-
-    // Assumes store object is always the same
-    this.store = store
 
     if (!this.isRouteWithoutStore) await this.fetchUser()
   }
@@ -163,12 +161,12 @@ class App extends NextApp {
     return null
   }
 
-  render() {
+  AppShell = () => {
     const { store } = this
     const { Component, pageProps } = this.props
     const { isAuthorized } = this.state
 
-    if (!store || !isAuthorized) {
+    if (!isAuthorized) {
       return (
         <>
           <Application dataProvider={undefined} activity="overview" />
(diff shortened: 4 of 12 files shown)


components/forms/change-password/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <Card className={className} tag={tag}>
+      <h2>Change password</h2>
+      <iframe
+        ref={iframeRef}
+        className={styles.changePasswordIframe}
+        title="Password change form"
+        src={`/secure/reset.html?${search}`}
+        scrolling="no"
+        allowtransparency="true"
```
viktor-yakubiv, line 36: Why is this needed? It's not standard but the default behaviour is that the iframe is transparent.

components/forms/change-password/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <Card className={className} tag={tag}>
+      <h2>Change password</h2>
+      <iframe
+        ref={iframeRef}
+        className={styles.changePasswordIframe}
+        title="Password change form"
+        src={`/secure/reset.html?${search}`}
+        scrolling="no"
```
viktor-yakubiv, line 35: This attribute is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#Attributes).

components/forms/change-password/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+        allowtransparency="true"
+        onLoad={() => {
+          const observer = new MutationObserver(resize)
+          observer.observe(iframeRef.current.contentWindow.document.body, {
+            attributes: true,
+            attributeOldValue: false,
+            characterData: true,
+            characterDataOldValue: false,
+            childList: true,
+            subtree: true,
```
viktor-yakubiv, line 45: Do not observe more than you actually need.

components/forms/change-password/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+      <iframe
+        ref={iframeRef}
+        className={styles.changePasswordIframe}
+        title="Password change form"
+        src={`/secure/reset.html?${search}`}
+        scrolling="no"
+        allowtransparency="true"
+        onLoad={() => {
+          const observer = new MutationObserver(resize)
+          observer.observe(iframeRef.current.contentWindow.document.body, {
```
viktor-yakubiv, line 39: Form size never changes. Why do you observe it?

components/forms/change-password/styles.css
```diff
@@ -0,0 +1,5 @@
+.change-password-iframe {
+  width: 100%;
+  max-width: 50rem;
+  height: 100%;
```
viktor-yakubiv, line 4: I am not sure `height` property is applied.

components/forms/index.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export { default as ChangePassword } from './change-password'
```
viktor-yakubiv, line 2: You could use more modern syntax, we have a plugin for it.

```suggestion
export ChangePassword from './change-password'
```

pages/_app/index.jsx
```diff
@@ -28,11 +28,14 @@ class App extends NextApp {
...
 
+  store = initStore()
+
   handleRouteChange = (url) => {
     logPageView(url)
     const { dataProvider, activity } = new Route(url)
 
-    if (dataProvider) {
+    // eslint-disable-next-line no-restricted-globals
+    if (!isNaN(dataProvider)) {
```
viktor-yakubiv, line 38: What the reason to use global instead Number?

If you ever have to disable linter, always add a comment before with the explanation of the reason for doing so.

pages/_app/index.jsx
```diff
@@ -28,11 +28,14 @@ class App extends NextApp {
     isAuthorized: false,
   }
 
+  store = initStore()
```
viktor-yakubiv, line 31: It executes on the server, does not it? Should it be so?

pages/_app/index.jsx
```diff
@@ -164,12 +163,12 @@ class App extends NextApp {
     return null
   }
 
-  render() {
+  AppShell = () => {
```
viktor-yakubiv, line 166: It seems to be an antipattern to me. Why do you need this?

pages/data-providers/[data-provider-id]/settings.css
```diff
@@ -4,4 +4,8 @@
 
 .section {
   margin-bottom: calc(1.5 * var(--component-padding-y, 1rem));
+
+  & input {
+    max-width: 50rem;
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/settings.css:9]: You use this value twice. Is not it better to apply this to form as a class and then apply it the iframe.

public/secure/reset.js
```diff
@@ -0,0 +1,104 @@
...
+    credentials: 'include',
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded',
+    },
+  }).then((response) => {
+    if (response.status === 401)
+      throw new UnauthorisedError('Wrong credentials')
+
+    // TODO: We should distinguish errors.
+    //     Currently API sends only 400 `{"message": "error"}`
```
viktor-yakubiv, line 20: Yes, but the possible option here is only that the password and password repeat does not match. If something else happened, i.e. email mismatch the token, we should never know about. We just say: "Password reset error. Please request reset password again".

components/forms/change-password/index.jsx
Joozty: It changes when an error/success message is added

pages/_app/index.jsx
Joozty: server has always [new store](https://github.com/oacore/dashboard/blob/master/store/index.jsx#L15). So it should be safe.

pages/_app/index.jsx
Joozty: Actually this commit could be part of https://github.com/oacore/dashboard/pull/263. I need it there because I want to call [store method in Reset component](https://github.com/oacore/dashboard/blob/reset-password/pages/reset/step1.jsx#L20).

public/secure/reset.js
Joozty: I think 400 is sent instead of 401  when a cookie is not present,i.e. user accidentally logged out. This should be handled differently.

pages/_app/index.jsx
Joozty: Number.isNaN doesn't convert its argument to Number

> true if the given value is NaN and its type is Number; otherwise, false.

I will change it to Number.isNaN(Number(dataProvider))

pages/data-providers/[data-provider-id]/settings.css
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/settings.css:9]: it's going to be improved in https://github.com/oacore/dashboard/pull/263. I will leave it as is here for the time being.

public/secure/reset.js
viktor-yakubiv: You have a good point. @valerapro can you please do changes accordingly?

public/secure/reset.js
```diff
@@ -0,0 +1,114 @@
...
+const newPassword = document.getElementById('newPassword')
+const newPasswordAgain = document.getElementById('newPasswordAgain')
+
+function performApiRequest(url, data) {
+  return fetch(url, {
+    method: 'POST',
+    body: data,
+    credentials: 'include',
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded',
```
viktor-yakubiv, line 19: Is not this automatic?

public/secure/reset.js
```diff
@@ -0,0 +1,114 @@
...
+  else newPasswordAgain.setCustomValidity('')
+}
+
+function resetInputs() {
+  requiredInputs.forEach((el) => {
+    el.classList.remove('touched')
+  })
+}
+
+function changePassword(event) {
```
viktor-yakubiv, line 60: When you use const, there is no reason to use `function` unless you want to bind it or need a function prototype.

components/forms/change-password/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+    <Card className={className} tag={tag}>
+      <h2>Change password</h2>
+      <iframe
+        ref={iframeRef}
+        className={styles.changePasswordIframe}
+        title="Password change form"
+        src={`/secure/reset.html?${search}`}
+        onLoad={() => {
+          const observer = new ResizeObserver(resize)
+          observer.observe(iframeRef.current.contentWindow.document.body)
```
viktor-yakubiv, line 37: The observer never disconnects from the element. It's better to wrap it into an effect which you can dispose after unmounting.

components/forms/change-password/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+    email: store.user?.email,
+  }
+
+  // remove empty params
+  Object.keys(searchParams).forEach(
+    (key) => searchParams[key] === undefined && delete searchParams[key]
+  )
+
+  const search = new URLSearchParams(searchParams).toString()
+  const resize = () => {
```
viktor-yakubiv, line 23: I would use a callback hook but it's minor.

components/forms/change-password/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+    iframeRef.current.style.height = `${iframeRef.current.contentWindow.document.body.offsetHeight}px`
```
viktor-yakubiv, line 24: Does not it work without this hack?

components/forms/change-password/index.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { useRef } from 'react'
+
+import styles from './styles.css'
+
+import { Card } from 'design'
+import { withGlobalStore } from 'store'
+
+const { IDP_URL } = process.env
+
+const ChangePassword = ({ className, store, tag = 'section' }) => {
```
viktor-yakubiv, line 10: I do not think makind it being section here is a good idea. You could pass `className` and `tag` as rest props . Also, there is no reason to deny the user pass other HTML attributes.

components/forms/change-password/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+  const iframeRef = useRef(null)
+  const searchParams = {
+    identity_provider_url: IDP_URL,
+    email: store.user?.email,
+  }
+
+  // remove empty params
+  Object.keys(searchParams).forEach(
+    (key) => searchParams[key] === undefined && delete searchParams[key]
+  )
```
viktor-yakubiv, line 20: Try to avoid deleting props. It's not clear and does not perform very well.

```suggestion
  const email = store.user?.email
  const searchParams = {
    ...(email != null ? { email } : {}),
    identity_provider_url: IDP_URL,
  }
```

pages/data-providers/[data-provider-id]/settings.jsx
```diff
@@ -98,6 +98,8 @@ const Settings = ({ store, className, ...restProps }) => {
           tag="p"
         />
       </SettingsGroup>
+
+      <ChangePassword className={styles.section} />
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/settings.jsx:102]: Please, add `tag` here to make it consistent with other.

Joozty:
@viktor-yakubiv these so-called meaningless commits are for you so you don't need to go through all commits again and check if all of your suggestions from commits were fixed. 🙂 Why should I create 10 commits with one line fixes? You won't make me sorry.

components/forms/change-password/index.jsx
Joozty: No, it doesn't...

public/secure/reset.js
Joozty: Looks like the default value is `Content-Type: application/json` However @valerapro is working on accepting JSON. So I will adjust this code once it's done.

viktor-yakubiv:
I do not mind how you do your commits during the development/review/development but I care what we have in the end. I recommend you to resolve your commits accordingly to the tasks you tried to solve, so in the end, you know that you squash this one with the first, and that one with the second commit. So before merging you can do it and have a well-described job.

"Fixes after review" does not mean anything in the `master` branch and does not mean anything in any squashed commits. I was only to improve existing chunks of job.

public/secure/reset.js
viktor-yakubiv: It's not. I guess it's empty or `text/plain` by default. Since fetch is a low-level API, it should be empty by default.

Joozty: I tried it without and it gave me that header value.

components/forms/change-password/index.jsx
Joozty: There is no need to pass other HTML attributes. If there is we can simply add it.

PR #259, opened 2020-04-14
Move login to /login

Joozty:
@viktor-yakubiv I know you will hate me for several reasons.

1. Most of the work is in the first commit. Most of the things were coupled together so I had to do it like that and also I was a little bit lazy (won't lie to you) because I needed to spare time for reset/change feature
2. Added next/link. This is the change you were against. I am not a fan of it either but we are now able to reflect dataProvider/activity in one place. We won't suffer from redirection loops and code splitting will be functional.
3. Finally, login has its own URL. At least one positive thing for you. 
4. The initial loading animation was removed. I don't think we suffer from slow user endpoint now and it makes the decoupling easier.
5. All login assets were moved to /public/secure

Updated:

Looks like user endpoint is slow on production. Maybe I should add animation back or at least add there the loading bar? What do you think?


Joozty:
Go ahead. I am really happy that you looked at this PR with priority because it kind of blocks my Change Password PR.

pages/data-providers/[data-provider-id]/plugins/index.jsx
```diff
@@ -1,25 +1,28 @@
...
+        href={`/data-providers/[data-provider-id]/${href}`}
+        as={`/data-providers/${store.dataProvider.id}/${href}`}
+      >
+        <Button variant="contained" href={href} tag="a">
+          {actionCaption}
+        </Button>
+      </Link>
+    </Card>
+  )
 )
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/plugins/index.jsx:26]: This was a pure component, independent from 'store'. Now it's tightly coupled

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/plugins/index.jsx:26]: Relative links don't work. I can try to import the store directly and don't observe this component at all but it's basically the same.

Joozty:
🤞 I hope I haven't caused more 🐛 with this PR.

PR #262, opened 2020-04-14
Handle login in iframe only

Joozty:
The title says everything I guess. The app doesn't communicate with iframe at all now.

Nice to have in the future:
We should probably show some placeholder cards while waiting for /user response. It takes some time. This could make a smoother UI.

### DIFF
--- a/public/secure/login.js
+++ b/public/secure/login.js
@@ -27,10 +27,6 @@ function removeLoadingAnimation() {
   form.classList.remove('loading')
 }
 
-function handlePostMessage(event) {
-  if (event.data === 'login-finished') removeLoadingAnimation()
-}
-
 function showWrongCredentialsMessage() {
   const message = document.getElementById('message')
   message.innerHTML = 'The username or password you entered is incorrect.'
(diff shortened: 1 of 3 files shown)


public/secure/login.js
```diff
@@ -57,28 +53,23 @@ function login(event) {
...
-            () => 'login-fallback',
-            () => 'login-fallback'
-          )
-        }
-
-        if (error instanceof UnauthorisedError) showWrongCredentialsMessage()
-        return 'login-processing'
+  performLoginRequest(loginUrl, data).then(
+    () => {
+      window.top.location.href = '/'
```
viktor-yakubiv, line 58: Why `.top`?

This line breaks the architectural principle of self-containing. If it happens somehow that your parent app decides to host within some path, let's say `core.ac.uk/dashboard/`, this will break. Technically it should either accept a parameter or send a notification what it was before.

Joozty: > Why .top?

I found it on stackoverflow so I used it. There is no big difference betweenthese three I am aware of.
```
window.parent.window.location.href
parent.location.href
window.top
```


That's true but we would need to handle this everywhere (logout, actions...) not just here.

PR #263, opened 2020-04-15
Add password reset forms

Joozty:
@viktor-yakubiv Feel free to do early review. I haven't had a chance to test it but in theory, reset email will contain this link: `http://dashboard.url/reset/step2?token=<token_here>&email=<email_here>`

For dev api dashhoard url will be replaced by origin header
for prod api dashboard url will be replaced by beta.dashboard.core.ac.uk

### DIFF
--- a/pages/reset/[token].jsx
+++ b/pages/reset/[token].jsx
@@ -1,0 +1,17 @@
+import React from 'react'
+import { withRouter } from 'next/router'
+
+import styles from './styles.module.css'
+
+import { ChangePassword } from 'components/forms'
+
+const CreatePassword = ({ router }) => (
+  <ChangePassword
+    tag="div"
+    className={styles.resetContainer}
+    email={router.query.email}
+    token={router.query.token}
+  />
+)
+
+export default withRouter(CreatePassword)
--- a/pages/reset/index.jsx
+++ b/pages/reset/index.jsx
@@ -1,0 +1,43 @@
+import React, { useState } from 'react'
+import { classNames } from '@oacore/design/lib/utils'
+
+import styles from './styles.module.css'
+
+import { Card, TextField, Button } from 'design'
+import { withGlobalStore } from 'store'
+
+const ResetPassword = ({ className, store, ...restProps }) => {
+  const [message, setMessage] = useState(null)
+
+  // TODO: Better texts and move it to @oacore/texts
+  const handleSubmit = async (event) => {
+    event.preventDefault()
+    try {
+      const formData = new FormData(event.target)
+      const data = {
+        email: formData.get('email'),
+      }
+      await store.user.requestResetToken(data)
+      setMessage('Please check your mailbox with the following instructions.')
+    } catch (error) {
+      setMessage("Something wen't wrong. Please try it again!")
+    }
+    return false
+  }
+
+  return (
+    <Card
+      className={classNames.use(styles.resetContainer, className)}
+      {...restProps}
+    >
+      <form onSubmit={handleSubmit}>
+        <h2>Reset password</h2>
+        <TextField id="email" label="Email" name="email" tag="p" required />
+        <Button variant="contained">Continue</Button>
+        {message && <p>{message}</p>}
+      </form>
+    </Card>
+  )
+}
+
+export default withGlobalStore(ResetPassword)
--- a/public/secure/reset.js
+++ b/public/secure/reset.js
@@ -7,6 +7,7 @@ const email = document.getElementById('email')
 const token = document.getElementById('token')
 const form = document.getElementById('password-form')
 const requiredInputs = form.querySelectorAll('[required]')
+const password = document.getElementById('password')
 const newPassword = document.getElementById('newPassword')
 const newPasswordAgain = document.getElementById('newPasswordAgain')
 
@@ -23,7 +24,7 @@ function performApiRequest(url, data) {
     //     Currently API sends only 400 `{"message": "error"}`
     if (response.status !== 200) {
       throw new WrongInputError(
-        'Looks like you typed in wrong password. Please try it again!'
+        'Password reset error. Please request reset password again!'
       )
     }
     return response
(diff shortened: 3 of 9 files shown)


pages/_app/index.jsx
```diff
@@ -21,7 +21,7 @@ process.on('uncaughtException', (err) => {
   Sentry.captureException(err)
 })
 
-const ROUTES_WITHOUT_STORE = ['/login']
+const ROUTES_WITHOUT_STORE = ['/login', '/reset']
```
viktor-yakubiv, line 24: Buy the way, why do you need this? Is it bad if you pass store to the Component but do not use it there?

pages/reset/index.js
```diff
@@ -0,0 +1,11 @@
+import { useEffect } from 'react'
+import Router from 'next/router'
+
+const Index = () => {
+  useEffect(() => {
+    Router.push('/reset/step1')
```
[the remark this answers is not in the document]
Joozty: I didn't think about it too much. I implemented first idea which came to my mind. It can be anything. I would personally go for `dashboard.core.ac.uk/reset?verify={token}`...

pages/_app/index.jsx
Joozty: it's not about the store. Pages specified in this constant don't require to fetch user at componentDidMount, i.e. it avoids redirecting to /login page when you are on /reset.

viktor-yakubiv: We should handle it differently but not at the moment. Thank you for the explanation!

pages/reset/index.js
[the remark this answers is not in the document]
viktor-yakubiv: Both are fine to me. It depends on how you are going to handle stuff: in a single file or using 2 ones.

pages/reset/step1.jsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <Card
+      className={classNames.use(styles.resetContainer, className)}
+      {...restProps}
+    >
+      <form onSubmit={handleSubmit}>
+        <h2>Reset password - step 1</h2>
+        <TextField id="email" label="Email" name="email" tag="p" required />
+        <Button variant="contained">Reset</Button>
+        <div>{message}</div>
```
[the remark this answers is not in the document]
viktor-yakubiv: It's minor, sorry

pages/reset/index.js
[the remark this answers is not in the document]
Joozty: dashboard.core.ac.uk/reset/{token} it is then.

@valerapro please can you update reset link to the following URL:

 dashboard.core.ac.uk/reset/{token}?email="email@here.com"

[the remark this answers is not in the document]
viktor-yakubiv: You should send the link along with the rest request. The token is fulfilled by the placeholder replacement and inserted to email.

pages/reset/[token].jsx
```diff
@@ -0,0 +1,17 @@
...
+const ResetStep2 = ({ router }) => (
+  <ChangePassword
+    tag="div"
+    className={styles.resetContainer}
+    email={router.query.email}
+    token={router.query.token}
+  />
+)
+
+export default withRouter(ResetStep2)
```
Comment (viktor-yakubiv) [inline pages/reset/[token].jsx:17]: I am struggling to find a better name but the following is the best I can imagine at the moment. When numbers appear in the name it's a sign that something is not good.

```suggestion
const CreatePassword = ({ router }) => (
  <ChangePassword
    tag="div"
    className={styles.resetContainer}
    email={router.query.email}
    token={router.query.token}
  />
)

export default withRouter(CreatePassword)
```

Why do not you use `useRouter` hook instead? Just curious 🧐 Your option maybe more flexible

pages/reset/[token].jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { withRouter } from 'next/router'
+
+import styles from './styles.module.css'
+
+import { ChangePassword } from 'components/forms'
+
+const ResetStep2 = ({ router }) => (
+  <ChangePassword
```
Comment (viktor-yakubiv) [inline pages/reset/[token].jsx:9]: Do not you want to have a heading somewhere?

pages/reset/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+    event.preventDefault()
+    try {
+      const formData = new FormData(event.target)
+      const data = {
+        email: formData.get('email'),
+      }
+      await store.user.requestResetToken(data)
+      setMessage('Please check your email for following instructions.')
+    } catch (error) {
+      setMessage("Something wen't wrong. Please try it again!")
```
viktor-yakubiv, line 23: I am not sure it's good to use exclamation mark here. @mcancellieri do you have any ideas?

public/secure/reset.js
```diff
@@ -23,7 +24,7 @@ function performApiRequest(url, data) {
     //     Currently API sends only 400 `{"message": "error"}`
     if (response.status !== 200) {
       throw new WrongInputError(
-        'Looks like you typed in wrong password. Please try it again!'
+        'Password reset error. Please request reset password again!'
```
viktor-yakubiv, line 27: The same here about the exclamation mark.

I would slightly change the message but I do not know what exactly it should be. Leave it as is, I will review it with @nancypontika when have time.

pages/reset/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+  // TODO: Better texts and move it to @oacore/texts
+  const handleSubmit = async (event) => {
+    event.preventDefault()
+    try {
+      const formData = new FormData(event.target)
+      const data = {
+        email: formData.get('email'),
+      }
+      await store.user.requestResetToken(data)
+      setMessage('Please check your email for following instructions.')
```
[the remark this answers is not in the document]
Joozty: These phrases need to be definitely reviewed.

pages/reset/[token].jsx
Comment (Joozty) [reply pages/reset/[token].jsx:17]: No reason I just forgot that it existed.

pages/reset/[token].jsx
Comment (Joozty) [reply pages/reset/[token].jsx:9]: What would you put there?

Comment (viktor-yakubiv) [reply pages/reset/[token].jsx:9]: Like _Create new password_ maybe

Comment (Joozty) [reply pages/reset/[token].jsx:9]: Actually there is a headline inside that component. it's shared with settings section https://dashboard-git-reset-password.oacore.now.sh/reset/4545

pages/reset/index.jsx
Joozty: I didn't like it too but using dot is not correct too I guess. As I said these phrases should be carefully reviewed and changed at some point.

PR #265, opened 2020-04-15
Upgrade @oacore/design to 0.1.0, add library configuration

viktor-yakubiv:
Upgrades @oacore/design to version 0.1.0.

Adds `design.config.js` with the icon mapping needed in the project. The configuration uses some extra icons needed in the future.

Adds `@mdi/svg` as a dependency to extract and build all necessary icons.

Replaces icon names to the proper ones from the `@mdi/svg` package. Changes URLs to shorthand syntax to freshly introduced one (start from `#` to use the sprite).

---

@Joozty since `@oacore/design` supports configuration files kind of `design.config.js` it adds such config to build all needed icons at the moment right into the project public directory. It adds some icons for the future, what I personally do not like but I assume we will have it when this is merged. All icons for the Reader are removed.

Also, it resolves most of the breaking changes introduced by using icons from `@mdi/svg`. Although, I might miss anything so please review it and let me know.

~It depends on oacore/design#166 and **must wait** for the next release of the package.~ Since we released the new version, this one is ready.

Please, be aware you will need to do the same for in the oacore/reader to use the new version.

### DIFF
--- a/components/application/activity-select/index.jsx
+++ b/components/application/activity-select/index.jsx
@@ -18,7 +18,7 @@ const linkProps = (value, dataProvider) => {
     as: `/data-providers/${dataProvider}/${value}`,
   }
 }
-const toIcon = (value) => `/design/icons.svg#${activities.get(value).icon}`
+const toIcon = (value) => `#${activities.get(value).icon}`
 
 const ActivitySelect = ({ children }) => (
   <Drawer.List className={styles.list}>{children}</Drawer.List>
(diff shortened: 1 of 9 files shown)


viktor-yakubiv:
This works like a charm. Almost. The `postinstall` is triggered only on a clean install, i.e. first time with a fresh project and each time on `npm ci`. I hope, it does not break production builds. If you need to update something during the development, you can easily use `npx @oacore/design build`, it will rebuild icons and runtime config.

For the future, it would be nice to add automatic global CSS copying as a `copy` command but I am not sure in this. However, it would be really good to have `clean` command that removes icons from the project.

---

I updated [my comment above](https://github.com/oacore/dashboard/pull/265#issue-403777104) accordingly to the commit message. @Joozty could you please review?

PR #266, opened 2020-04-16
Use css from design in login iframe

Joozty:


form/iframe-form.jsx
```diff
@@ -0,0 +1,24 @@
...
+  const ref = useRef(null)
+  const resize = useCallback(() => {
+    ref.current.style.height = `${ref.current.contentWindow.document.body.offsetHeight}px`
+  }, [])
+
+  useEffect(() => {
+    const observer = new ResizeObserver(resize)
+    observer.observe(ref.current.contentWindow.document.body)
+    return () => observer.disconnect()
+  }, [])
```
viktor-yakubiv, line 15: Should not this depend on `ref` or `ref.current`?

Joozty: ref is created only once.

PR #268, opened 2020-04-17
New time lag chart  without aggregation

Joozty:


viktor-yakubiv:
Apollo data is not good. I saw it but forgot about, sorry. We should probably use a logarithmic Y scale to avoid 1-pixel heigh columns. Also, the chart is extremely long.

Joozty:
@viktor-yakubiv I decided to give a [fixed size](https://github.com/oacore/dashboard/blob/time-lag-chart-new/components/time-lag-chart/index.jsx#L115) to every bar. We can limit max-width if you want. Something like

```js
   width={Math.min(normalizedData.length * 10, 5000)} // max 5000px
```

viktor-yakubiv:
The bar width can be smaller. 4px should be fine I think. In the design, I used 4px with 2px gap

viktor-yakubiv:
This may be helpful for you. It's not tested but I am sure you can easily get what it does

```js
const useBoundary = (ref, options = {}) => {
  const {
    initialShift, // index of the column 0, initial shift
    barCount = 92,
    barWidth = 4,
    gutterWidth = 2,
  } = options

  const { floor } = Math

  const element = useRef(ref).current

  useEffect(() => {
    const { clientWidth } = element
    const barVisibleCount = Math.floor(
      (clientWidth + gutterWidth) / (barWidth + gutterWidth)
    )

    const startBarIndex = barVisibleCount > barCount
      ? initialShift - Math.floor((barsPerClientWidth - barCount) / 2)
      : initialShift
    const scrollPosition =
      barWidth * (startBarIndex - 1) +
      gutterWidth * (startBarIndex - 1)

    element.scroll(scrollPosition, 0)
  }, [])
}
```

Joozty:
@viktor-yakubiv Sorry I was not able to make it work according to your example. I was kind of lost in those calculations and I would probably need to add min-width to chart wrapper or fire that hook somehow clever (with mutation observer). However, if you can make it work feel free to push here. 🙂

viktor-yakubiv:
I think the data change only once, does not it? We should render some loading component if we do not have the data and rerender with chart component once we have it. In such case it should work.

Joozty:
That's true. However, what do you expect from this? Code would more readable for you? Obviously I expect it should be faster but on the other hand it's fired only. Also from a number of code lines, it's almost identical.

PR #269, opened 2020-04-20
Add Cross Repository Check card

Joozty:


Joozty:
@viktor-yakubiv I am not sure if I used correct values from API. Can you please double-check. Also, there is one TODO comment about "days value". I am not sure where to get it from.

viktor-yakubiv:
The days is the minimum of the all papers deposited earlier in other repositories.

pages/data-providers/[data-provider-id]/deposit-dates.jsx
```diff
@@ -62,23 +66,23 @@ const DepositDates = ({
...
+      <Card.Title tag="h2">{texts.crossRepositoryCheck.title}</Card.Title>
+      <NumericValue
+        value={valueOrDefault(
+          store.depositDates.crossDepositLag?.possibleBonusCount,
+          'Loading...'
+        )}
+        caption="outputs match"
+        className={styles.outputsMatch}
+      />
+      <Button tag="a" href="#crossRepositoryCheck" variant="contained">
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-dates.jsx:79]: As I wrote somewhere, I prefer hyphenized IDs because it's more like HTML and URL.
```suggestion
      <Button variant="contained" href="#cross-repository-check" tag="a" >
```

pages/data-providers/[data-provider-id]/deposit-dates.jsx
```diff
@@ -91,6 +95,27 @@ const DepositDates = ({
...
+        {texts.crossRepositoryCheck.description}
+      </Card.Description>
+      <p>
+        {store.depositDates.crossDepositLag
+          ? texts.crossRepositoryCheck.body.render({
+              nonCompliantCount:
+                store.depositDates.crossDepositLag.nonCompliantCount,
+              recordsInAnotherRepository:
+                store.depositDates.crossDepositLag.possibleBonusCount,
+              days: '#', // TODO: Where to get this value?
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-dates.jsx:110]: Maybe we should not have this before we have actual table data.

texts/depositing/index.js
```diff
@@ -27,5 +28,10 @@ const chart = {
   body: sourceChart.body,
 }
 
+const crossRepositoryCheck = {
+  ...sourceCrossRepositoryCheck.attributes,
+  body: new Template(sourceCrossRepositoryCheck.body),
```
viktor-yakubiv, line 33: We should improve our loader to simplify this.

pages/data-providers/[data-provider-id]/deposit-dates.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-dates.jsx:110]: You mean the whole card or only the last few words?

Joozty:
Can we please cover 0 in new PR. It's really hard to keep track of all PRs and resolving conflicts. It is driving me crazy...

pages/data-providers/[data-provider-id]/deposit-dates.jsx
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/deposit-dates.jsx:109]: I mean exactly `days`.

viktor-yakubiv:
Yes, sure. Please, resolve what you can here and move 0-case to a separate PR. Thank you 👍

PR #270, opened 2020-04-20
Compliance data overview card

Joozty:


components/numeric-value/index.jsx
```diff
@@ -26,8 +26,15 @@ const formatNumber = (
...
 }
 
-const NumericValue = ({ value, append, caption, diff, ...restProps }) => (
-  <Numeral tag="p" {...restProps}>
+const NumericValue = ({
+  value,
+  append,
+  caption,
+  diff,
+  tag = 'p',
```
viktor-yakubiv, line 34: Why do you override default? It seems the default should be `p`

pages/data-providers/[data-provider-id]/compliance.jsx
```diff
@@ -40,26 +40,31 @@ const SidebarContent = ({ context: { oai, originalId, authors, title } }) => {
...
-const DepositDates = ({
-  className,
-  store,
-  tag: Tag = 'main',
-  ...restProps
-}) => (
+const Compliance = ({ className, store, tag: Tag = 'main', ...restProps }) => (
   <Tag className={[styles.container, className].join(' ')} {...restProps}>
-    <h1>Deposit compliance</h1>
+    <h1>REF 2021 Compliance</h1>
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/compliance.jsx:45]: Sorry, it was a mistake. I think we should not change this line without proper discussion with the team. Can you please leave this line unchanged?

pages/data-providers/[data-provider-id]/compliance.jsx
```diff
@@ -40,26 +40,31 @@ const SidebarContent = ({ context: { oai, originalId, authors, title } }) => {
...
 
-    <Card className={styles.complianceLevel} tag="section">
+    <Card className={styles.dataOverview} tag="section">
       <h2>{texts.compliance.title}</h2>
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/compliance.jsx:48]: If you change the scope it would be good to change it everywhere. In `texts` case may be better to create a new one.

pages/data-providers/[data-provider-id]/compliance.css
```diff
@@ -15,9 +15,18 @@
   }
 }
 
-.compliance-level {
+.data-overview {
   flex: 2 2 60%;
   width: 100%;
+
+  & > .numbers {
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/compliance.css:22]: There is no reason to use nested class. It performs worse. It matters even if we talk about such simple one because we have a lot of selectors.

pages/data-providers/[data-provider-id]/compliance.css
```diff
@@ -15,9 +15,18 @@
...
+    display: flex;
+    flex-wrap: wrap;
+    align-items: flex-end;
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/compliance.css:25]: I am not sure we should align numbers to the bottom; `baseline` or `flex-start` should be fine. What do you try to cover by this line?

pages/data-providers/[data-provider-id]/compliance.jsx
```diff
@@ -40,26 +40,31 @@ const SidebarContent = ({ context: { oai, originalId, authors, title } }) => {
...
-          {texts.compliance.body.render({
-            amount: (100 - store.depositDates.complianceLevel).toFixed(1),
-          })}
-        </Markdown>
-      ) : (
-        <p>{texts.noData.body}</p>
-      )}
+      <div className={styles.numbers}>
+        <NumericValue
+          tag="div"
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/compliance.jsx:51]: `p` is fine here, a [paragraph](https://html.spec.whatwg.org/#paragraph) despite it's commonly a few sentences of text in the web it is a small chunk of self-contained information and address, fieldset or a number + its description can be wrapped into `p`.

pages/data-providers/[data-provider-id]/compliance.jsx
```diff
@@ -40,26 +40,31 @@ const SidebarContent = ({ context: { oai, originalId, authors, title } }) => {
...
+          caption="outputs counted"
+        />
+        <NumericValue
+          tag="div"
+          value={valueOrDefault('2.7', 'Loading...')}
+          append="%"
+          caption="non-compliant"
+        />
+      </div>
+      <Button tag="a" href="#depositDatesCard" variant="contained">
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/compliance.jsx:65]: I prefer URLs, IDs, classes with _kebab-case_, JS/Java names with _camelCase_ (along with Python/C++ names in _snake_case_). It's the way a certain language or domain tends to look like. We follow this everywhere. Could you please change it accordingly here?

We should standardize this properly organisation-wise.

texts/depositing/compliance-notice.md
```diff
@@ -1,9 +1,5 @@
 ---
-title: Compliance
+title: Data Overview
```
viktor-yakubiv, line 2: Americans prefer _Data Overview_ (each word capital).  
British prefer _Data overview_ (sentence case).

CORE is where knights and queen, we prefer British.

store/activities.js
```diff
@@ -10,8 +10,8 @@ const config = [
     icon: 'document',
   },
   {
-    id: 'depositDates',
-    path: 'deposit-dates',
+    id: 'compliance',
+    path: 'compliance',
```
viktor-yakubiv, line 14: _Compliance_ is not clear enough, it's too generic, compliance can be inspected comparingly to anything. I would keep it `deposit-compliance`.

texts/depositing/compliance-notice.md
```diff
@@ -1,9 +1,5 @@
...
+redirect: Review data
 ---
 
-You have **{{amount}}%** outputs are non-compliant according to
```
viktor-yakubiv, line 5: As I wrote above, please create a separate text scope for this change. We louse this information at the moment but we should not. It should stay on the page.

I would display this below the page title in grey text. Let me know if you need designer help with this.

utils/helpers.js
```diff
@@ -1,6 +1,10 @@
 export const range = (length, start = 0, step = 1) =>
   Array.from(Array(length), (_, i) => start + i * step)
```
viktor-yakubiv, line 2: Such cool 🐍 function 🙂

texts/navigation.yml
```diff
@@ -4,7 +4,7 @@ items:
   index: Overview
   data: Data
   content: Content
-  deposit-dates: Deposit compliance
+  compliance: Compliance
```
viktor-yakubiv, line 7: Please, keep this as it was because it was not discussed with the team.

I used shorter word on the mockup because I would need to change the sidebar size on every mockup in the project otherwise.

Joozty: I thought you discussed this with David since it was in design

Will revert it.

pages/data-providers/[data-provider-id]/compliance.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/compliance.jsx:51]: I didn't want to remove all padding and margin stuff from paragraph hence `div`

texts/depositing/compliance-notice.md
Joozty: Please let me know where shoudl this card be placed or add it on your own. Whatever makes it faster. :)

Joozty: Sorry I missed the second paragraph. I will add it below the title.

mcancellieri:
It's a bit of a tricky point because we will need to run practically a full dump for it.
If you get the percentage shouldn't you also have the number of outputs from the same place?

Joozty:
I am not sure about the data overview card. I think it should show more data for >fullhd screens otherwise there is a big gap

dataverview

viktor-yakubiv:
There should be 2 equal columns with text aligned to the left. It will be more-or-less balanced and satisfies me enough.

However, if you have any suggestion about how to make it look better, I would really appreciate it.

pages/data-providers/[data-provider-id]/deposit-compliance.css
```diff
@@ -15,11 +15,19 @@
...
-.compliance-level {
+.data-overview {
   flex: 2 2 60%;
   width: 100%;
 }
 
+.numbers {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-compliance.css:26]: As I already said, convert child elements to proper columns with equal size.

pages/data-providers/[data-provider-id]/deposit-compliance.jsx
```diff
@@ -40,33 +40,50 @@ const SidebarContent = ({ context: { oai, originalId, authors, title } }) => {
...
-          })}
-        </Markdown>
-      ) : (
-        <p>{texts.noData.body}</p>
-      )}
+    <Card className={styles.dataOverview} tag="section">
+      <h2>{texts.dataOverview.title}</h2>
+      <div className={styles.numbers}>
+        <NumericValue
+          tag="div"
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-compliance.jsx:63]: I still would use `p` here as it represents a self-contained piece of information that sounds exactly like _2,043 outputs counted_, it's a sentence.

pages/data-providers/[data-provider-id]/deposit-compliance.jsx
```diff
@@ -40,33 +40,50 @@ const SidebarContent = ({ context: { oai, originalId, authors, title } }) => {
...
+        {texts.dataOverview.redirect}
+      </Button>
     </Card>
 
     <Card className={styles.export} tag="section">
       <h2>{texts.exporting.title}</h2>
       <p>
         {texts.exporting.description.render({
-          count: store.depositDates.depositDatesCount || '',
+          count: store.depositDates.complianceLevel[0] || '',
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-compliance.jsx:86]: ~Why it's `[0]` here. It's not really representative.~ Check the comment in `store`.

texts/depositing/data-overview.md
```diff
@@ -0,0 +1,4 @@
+---
+title: Data Overview
```
viktor-yakubiv, line 2: British-like title in sentence case.

```suggestion
title: Data overview
```

texts/depositing/compliance-notice.md
viktor-yakubiv: This is not what I meant. Could you please rever all changes in this card and keep it unused at the moment?

store/deposit-dates.js
```diff
@@ -43,7 +40,7 @@ class DepositDates extends Store {
 
     if (total === 0) return 0
     const level = (compliant / total) * 100
-    return Math.round(level * 100) / 100
+    return [total, Math.round(level * 100) / 100]
```
viktor-yakubiv, line 43: I would create 3 separate computed properties: `totalCount`, `nonCompliantCount`, `complianceLevel`. It's more clear than `complianceLevel[0]`

store/deposit-dates.js
```diff
@@ -34,25 +32,27 @@ class DepositDates extends Store {
...
     this.getCrossDepositLag()
   }
 
+  @computed
+  get totalCount() {
+    return this.timeLagData.reduce((acc, curr) => acc + curr.worksCount, 0)
+  }
+
+  @computed
+  get compliantCount() {
```
viktor-yakubiv, line 45: I would really leave `nonCompliantCount` because it's logically more correct.

store/deposit-dates.js
```diff
@@ -34,25 +32,27 @@ class DepositDates extends Store {
...
-        if (curr.depositTimeLag <= 90) acc[1] += curr.worksCount
-        return acc
-      },
-      [0, 0]
-    )
-
-    if (total === 0) return 0
-    const level = (compliant / total) * 100
+    if (this.totalCount === 0) return 0
+    const level = (this.compliantCount / this.totalCount) * 100
```
viktor-yakubiv, line 55: I am not sure this is better, a little hacky.

```suggestion
    const level = (this.compliantCount / (this.totalCount || 1)) * 100
```

pages/data-providers/[data-provider-id]/deposit-compliance.module.css
```diff
@@ -15,12 +15,24 @@
...
 }
 
+.numbers {
+  display: flex;
+  flex-wrap: wrap;
+  margin-bottom: 1rem;
+
+  & > p {
+    flex: 50%;
+    margin: 0;
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-compliance.module.css:30]: The following may not pass linter. I am trying to explicitly set flex parameters (grow is 0 by default) and add inter-column spacing just in case.

```suggestion
    flex: 1 0 50%;
    margin: 0 var(--component-padding-x, 1rem) var(--component-padding-y, 1rem);
```

pages/data-providers/[data-provider-id]/deposit-compliance.module.css
```diff
@@ -15,12 +15,24 @@
...
-.compliance-level {
+.data-overview {
   flex: 2 2 60%;
   width: 100%;
 }
 
+.numbers {
+  display: flex;
+  flex-wrap: wrap;
+  margin-bottom: 1rem;
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-compliance.module.css:26]: Fixes extra spacing below:
```suggestion
  margin: 0 calc(-1 * var(--component-padding-x, 1rem));
```

pages/data-providers/[data-provider-id]/deposit-compliance.jsx
```diff
@@ -52,17 +50,24 @@ const DepositDates = ({
...
-        <Markdown>
-          {texts.compliance.body.render({
-            amount: (100 - store.depositDates.complianceLevel).toFixed(1),
-          })}
-        </Markdown>
-      ) : (
-        <p>{texts.noData.body}</p>
-      )}
+    <Card className={styles.dataOverview} tag="section">
+      <h2>{texts.dataOverview.title}</h2>
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-compliance.jsx:54]: Why the `Card.Title` disappear? Is it because of some merge conflict?

pages/data-providers/[data-provider-id]/deposit-compliance.module.css
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-compliance.module.css:26]: I am not sure what you tried to achieve with this rule. It moves numbers to the left.

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/deposit-compliance.module.css:26]: The card has padding, let's say `16px`. I want to create 2 columns inside with 32px spacing between them.

To do so, I create a _row_ with a negative left and right margin (`-16px`) and 2 _columns_ inside the _row_ with `16px` left and right margin/padding.

It's a more flexible solution than using `:first-child` and `:last-child` with removing margin because of columns may break to the next row.

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-compliance.module.css]: it doesn't work when you make flex basis 50% and use wrap

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/deposit-compliance.module.css]: Yeah, the flex-basis should be relatively small. If it works with 1px, it can be so; otherwise any small amount that fits the number, e.g. 10rem.

PR #272, opened 2020-04-23
Add footer to table

Joozty:


components/table/styles.css
```diff
@@ -89,3 +72,23 @@
...
   color: var(--gray-400);
 }
+
+.footer {
+  background: var(--white);
+  border-top: 2px solid var(--table-border-color);
+}
+
+.footer-right {
+  float: right;
```
viktor-yakubiv, line 82: Should be flex in the future but this is fine too.

utils/helpers.js
```diff
@@ -1,6 +1,9 @@
 export const range = (length, start = 0, step = 1) =>
   Array.from(Array(length), (_, i) => start + i * step)
 
+export const formatNumber = (number) =>
+  new Intl.NumberFormat('en-GB', { maximumSignificantDigits: 3 }).format(number)
```
viktor-yakubiv, line 5: I am really happy you used `Intl.NumberFormat`, this is the most appropriate way.

I tested it, and I am not sure `maximumSignificantDigits` does exactly what you want. Just test it:

```js
new Intl.NumberFormat('en-GB', { maximumSignificantDigits: 3 }).format(123456.789) // outputs 123,000
```

I propose to make the function more configurable:

```suggestion
export const formatNumber = (number, { locale = 'en-GB', ...options } = {}) =>
  new Intl.NumberFormat(locale, options).format(number)
```

components/table/action.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { observer } from 'mobx-react'
+
+@observer
+class Action extends React.Component {
```
viktor-yakubiv, line 5: Why do you need empty observer wrapper?

Would not having observer component in place solve your problem?

```js
const TableButton = observer(ExportButton)
```

components/table/table.jsx
```diff
@@ -287,15 +273,30 @@ class InfiniteTable extends React.PureComponent {
...
-                    observe={showNextLoad}
-                    onVisible={this.loadNextPage}
-                  />
-                )}
                 {data && data.length === 0 && <NoDataFoundRow />}
               </Table.Body>
+              <tfoot className={styles.footer}>
+                <Table.Row>
+                  <Table.Cell colSpan={1000}>
+                    <div className={styles.footerLeft}>{action}</div>
```
viktor-yakubiv, line 281: Interesting solution 🤔

components/table/table.jsx
```diff
@@ -287,15 +273,30 @@ class InfiniteTable extends React.PureComponent {
...
+                    <div className={styles.footerRight}>
+                      {isFirstPageLoaded &&
+                        totalLength &&
+                        `Showing 1 - ${formatNumber(size)} of ${formatNumber(
```
viktor-yakubiv, line 285: I used [thin space](http://jkorpela.fi/chars/spaces.html) there but I am not sure in this.
```suggestion
                        // thin spaces between numbers
                        `Showing 1\u2009-\u2009${formatNumber(size)} of ${formatNumber(
```

components/table/action.jsx
Joozty: It was just copy-paste of the Column component. The observer is no needed there.

PR #273, opened 2020-04-24
Add publication dates chart

Joozty:


Joozty:
@viktor-yakubiv Don't you know if there is any way to add one ellipsis?

ellipsis

viktor-yakubiv:
Yes, I found a way to do it:

1. Remove `display: flex` from the bar
2. Convert spans to `inline-block`s
3. Add `float: left` to the first span
3. Add `text-align: right` to the bar
4. Remove whitespace and overflow properties from the second span, it's not needed
5. Add ~vertical~ padding to both spans to make the one is falling to the next line be hidden visually (you can add padding to the bar instead)

By the way, it's better to control height by the `label` and `value` (spans) or bar.

viktor-yakubiv:
Please use white text for the red bar.

Joozty:
@viktor-yakubiv great thanks, however, that doesn't give me ellipsis. Maybe I am doing something wrong? Would you mind to push your solution?

viktor-yakubiv:
I did, sorry. Add `max-width: 100%` for the floated element

Joozty:
You pushed it and I overrode it?

viktor-yakubiv:
I personally would prefer to keep `flex-basis: auto` (default) and set `width` (what would be the `flex-basis` for the flex element). This will allow you to use `min-width`

viktor-yakubiv:
Sorry, I did not push it (you would see a line above if I do so). I meant _I **did** forget_. I did not write any code, just played with it using developer tools.

Joozty:
Ok got it. Thanks for help :)

components/stacked-vertical-bar-chart/styles.css
```diff
@@ -4,21 +4,27 @@
...
   padding-right: 0.5rem;
   padding-left: 0.5rem;
   margin: 0;
   overflow: hidden;
+  text-align: right;
 }
 
 .bar > span {
+  display: inline-block;
+  padding: 0.25rem;
```
viktor-yakubiv, line 21: This may break If the line-height changes globally. I recommend to explicitly set `line-height: 1` here and change current padding to `0.5rem`.

pages/data-providers/[data-provider-id]/deposit-dates.jsx
```diff
@@ -91,6 +90,41 @@ const DepositDates = ({
...
+        <StackedVerticalBarChart
+          data={[
+            {
+              value: store.depositDates.publicationDatesValidate.fullCount,
+              caption: 'Matching',
+              background: 'var(--success)',
+            },
+            {
+              value: store.depositDates.publicationDatesValidate.partialCount,
+              caption: 'Incorrect',
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-dates.jsx:106]: The data here is actually **correct** but it's not complete, i.e only year matches. _Incomplete_ sounds slightly better. Maybe the best way to call it is _Partially matches_.

pages/data-providers/[data-provider-id]/deposit-dates.jsx
```diff
@@ -91,6 +90,41 @@ const DepositDates = ({
...
 
+    <Card tag="section">
+      <h2>Publication dates</h2>
+      <p>Compares publications dates with Crossref</p>
+      {store.depositDates.publicationDatesValidate ? (
+        <StackedVerticalBarChart
+          data={[
+            {
+              value: store.depositDates.publicationDatesValidate.fullCount,
+              caption: 'Matching',
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-dates.jsx:101]: Do you mind moving these messages to `texts`? It can be done later separately if you wish.

pages/data-providers/[data-provider-id]/deposit-dates.jsx
```diff
@@ -91,6 +90,41 @@ const DepositDates = ({
         )}
     </Card>
 
+    <Card tag="section">
+      <h2>Publication dates</h2>
+      <p>Compares publications dates with Crossref</p>
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-dates.jsx:95]: Does this look so because we do not have `Card.Description` yet?

components/stacked-vertical-bar-chart/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+const StackedVerticalBarChart = React.memo(({ data }) => {
+  const [transformedData, setTransformedData] = useState([])
+
+  useEffect(() => {
+    const total = data.reduce((acc, curr) => acc + curr.value, 0)
+    setTransformedData(
+      data.map((el) => ({ ...el, percentage: (el.value / total) * 100 }))
+    )
+  }, [data])
```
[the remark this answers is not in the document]
Joozty: it's a point of view. I took it from [here](https://github.com/oacore/dashboard/blob/master/components/enrichment-chart/index.jsx#L20). I don't mind both namings.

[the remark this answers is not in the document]
viktor-yakubiv: 🤦‍♂️

components/stacked-vertical-bar-chart/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+              title={`${caption} ${value}`}
+              className={styles.bar}
+              style={{
+                flexBasis: `${percentage}%`,
+                background,
+                color,
+              }}
+            >
+              <span className={styles.caption}>{caption}</span>
+              <span>{value}</span>
```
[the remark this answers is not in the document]
Joozty: I use type selector when I ran out of names ideas for classes 😄

pages/data-providers/[data-provider-id]/deposit-dates.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-dates.jsx:101]: I will do it later together with Card.Description

pages/data-providers/[data-provider-id]/deposit-dates.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-dates.jsx:95]: yep

pages/data-providers/[data-provider-id]/deposit-dates.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-dates.jsx:101]: Oh, it was already merged. Gonna do it here...

components/stacked-vertical-bar-chart/index.jsx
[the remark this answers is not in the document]
viktor-yakubiv: That's an option too 😄

PR #274, opened 2020-04-27
Remove unused getDerivedStateFromProps

Joozty:
I pushed unintentionally to master and noticed it just now. @viktor-yakubiv it was really small changes so I left it as is. Feel free to check them. This PR was supposed to contain those commits but it contains only the last piece.


viktor-yakubiv:
Thank you for the cleaning up.

PR #275, opened 2020-04-29
Use svg icons instead of css imitation

Joozty:


design/app-bar/toggle.module.css
```diff
@@ -2,11 +2,7 @@
...
-.bar {
-  display: block;
-  width: 24px;
-  height: 4px;
-  margin-bottom: 4px;
-  background: currentColor;
-  border-radius: 1px;
+.icon {
+  width: 2rem;
+  height: 2rem;
```
viktor-yakubiv, line 7: Do we need to resize the icon?

I'd prefer never resize it because it warps the image. Icons are drawn to be pixel perfect, they have 24x24 pixels size and each line is 2 pixels wide.

components/layout/side-bar.jsx
```diff
@@ -32,7 +32,12 @@ const SideBar = ({
           tag="button"
           onClick={() => dispatch({ type: 'toggle_sidebar' })}
         >
-          <CloseIcon aria-label="Close sidebar" />
+          <Icon
+            className={styles.icon}
+            src="#close"
+            aria-label="Close sidebar"
+            aria-hidden
```
viktor-yakubiv, line 39: I do not see really combination of `aria-hidden` and `aria-label`, these are mutually excluding.

The similar comment to `alt`. It looks silly but let's leave it.

```suggestion
            alt="Close icon"
            aria-label="Close sidebar"
```

PR #277, opened 2020-04-30
Use MobX react lite over mobx-react

Joozty:
I am not entirely sure about this. Currently, we don't observe any class so we can change it but I don't know about the future. On the other hand, it will force us to use more hooks and decomposite as much as possible to avoid big class components.

### DIFF
--- a/components/table/column.jsx
+++ b/components/table/column.jsx
@@ -1,7 +1,5 @@
 import React from 'react'
-import { observer } from 'mobx-react'
 
-@observer
 class Column extends React.Component {
   render() {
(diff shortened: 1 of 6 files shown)


viktor-yakubiv:
> On the other hand, it will force us to use more hooks 

I am not so sure about hooks because of tight coupling with React-ish approach.

> ... and decomposite as much as possible to avoid big class components.

However, this does not sound bad 😁

Joozty:
I heard rumors that React what to deliver class components in a separate package. Functional components should be "a new standard". Who knows what's true about it.

PR #281, opened 2020-04-30
Temporary fix now.sh deployment

Joozty:
For some strange reason sometimes deployment of branch fails (mostly on the first push). I think it's a Vercel issue. This is just the attempt to temporary fix it on our side.

https://vercel.com/oacore/dashboard/qm8ffsra7

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,6 @@
     "lint:css": "stylelint --ignore-path .gitignore '**/*.css'",
     "lint": "run-p lint:*",
     "postinstall": "run-p cp:*",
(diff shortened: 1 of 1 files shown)


viktor-yakubiv:
Why do you call 'temporary'

Joozty:
IMHO it should be fixed by Vercel guys. It worked before and that directory should exist before calling post-install script I assume. I mean I am not sure why sometimes deploy succeed and sometimes not. Maybe it will be a permanent fix. 😄

viktor-yakubiv:
> Maybe it will be a permanent fix. 😄

That is why I am against 'temporary' fixes. They are almost always 'permanent'

PR #285, opened 2020-04-30
Resource, User and store errors

viktor-yakubiv:
1. Creates a Resource the is a generic model for all API entities. The model can fetch itself from the API via URL and resolve nested scopes, i.e. can fetch itself and `dataProviders` (scope) if there is a `dataProvidersUrl`.
1. Creates User model based on the Resource
   1. Splits Dashboard-specific user from the generic User entity
   2. Removes old User entity
   3. Integrates new User entity into the Root
2. Adds Application Model errors
   - decouples API errors from the application model errors
   - uses new errors to handle Application issues
---

**⚠️ WARNING**: This is acceptable for the review but this is not tested due to API issues.


Joozty:
Feel free to continue. It looks good to me so far. 👍

viktor-yakubiv:
I faced a redirecting problem and have to hold this PR until it's fixed.

viktor-yakubiv:
@Joozty, unexpectedly, this PR started working right after the rebase and is ready to review.

viktor-yakubiv:
Also, the project builds and runs. I wonder if I need to check anything else 😄

viktor-yakubiv:
This has nothing to the data providers endpoint. Only User is slightly refactored.

I will check issues you reported.

viktor-yakubiv:
I rebased the branch and added a commit that fixes the error.

It's too early for our app to have anonymous access to the data. We need to do more refactoring to support this but I assume this PR is a good start.

pages/_app/index.jsx
```diff
@@ -68,7 +68,7 @@ class App extends NextApp {
   }
 
   handlePromiseRejection = (event) => {
-    if (event.reason instanceof UnauthorizedError) {
+    if (event.reason instanceof AuthorizationError) {
```
Joozty, line 71: there should be an or clause. If any request fails user should be immediately logged out. (The same applies for componentDidCatch method)


```
if (event.reason instanceof AuthorizationError || event.reason instanceof UnauthorizedError) {
}
```

viktor-yakubiv: Why do you think so?

Joozty: Because it's data protection also the dashboard may not work properly. For example, change password won't work and so on. Basically all protected endpoints will return error and user won't see any data. There is no notification that they have been logged out.

viktor-yakubiv: Thank you. I really appreciate you feedback. I will add changes appropriately.

PR #287, opened 2020-05-01
Add Issues tab

Joozty:


viktor-yakubiv:
I checked the preview. Despite I did not see any data, your design proposal is really good! I would not even think about this layout! 👍

viktor-yakubiv:
@Joozty what repository should I check? I cannot find any data

Joozty:
So dataprovider 5 should be good choice. Maybe ask Matteo to assign it to you. 

I haven't really tested last commit yet. Since API is hanging since the morning. So there may be some small bugs. :(

viktor-yakubiv:
I kind of finished and wonder if you dear @oacore/dashboard have any feedback. I am afraid we should preview it locally during these days because of deployment issues.

viktor-yakubiv:
@Joozty do you have time to help me solve conflicts? I assume it's because of the table changed.

Joozty:
Here is my approval. ✔️ Feel free to merge if you are not going to add anything else.

viktor-yakubiv:
There is an error with date formatting. I will handle it and merge one by one. I wonder if we should do an early release after this. @oacore/dashboard what do you think?

viktor-yakubiv:
I explored the error, it happens because of inconsistent [non-standard date formatting](https://stackoverflow.com/a/4310986/7205866) in the API response in Safari but is fine in Chrome and should be fine in Firefox.

Joozty:
Please report this to Jira or use dayjs if it helps. :)

PR #288, opened 2020-05-04
Fix card's headings

Joozty:


pages/data-providers/[data-provider-id]/settings.jsx
```diff
@@ -42,7 +42,7 @@ const Settings = ({ store, className, ...restProps }) => {
         className={classNames.use(styles.section).join(className)}
         tag="section"
       >
-        <h2>Organisation settings</h2>
+        <Card.Title tag="h2">Organisation settings</Card.Title>
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/settings.jsx:45]: It's weird. Does `Card.Title` apply other styles to the heading?

It should not, in my opinion. Also, it would be good to not apply `margin-bottom` to the title.

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/settings.jsx:45]: The reason why there I added there bottom margin is because when it is used with Card.Description there would be a too big gap. Because paragraph applies own margin-top.

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/settings.jsx:45]: I see. Thank you for the explanation.

PR #290, opened 2020-05-04
Show matching icon in deposit table

Joozty:


pages/data-providers/[data-provider-id]/deposit-compliance.jsx
```diff
@@ -36,6 +37,33 @@ const SidebarContent = ({ context: { oai, originalId, authors, title } }) => {
...
 
+class PublicationDateColumn extends Table.Column {
+  render() {
+    const {
+      context: { publicReleaseDate, publicationDateValidationLevel },
+    } = this.props
+
+    return (
+      <span>
+        {dayjs(publicReleaseDate).format('DD/MM/YYYY')}
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-compliance.jsx:48]: Can we utilise Intl here too? (for the future)

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-compliance.jsx:48]: What do you mean? The format is the same as if you would use Intl

```
console.log(new Intl.DateTimeFormat('en-GB').format(date));
// expected output: "20/12/2012"
```

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/deposit-compliance.jsx:48]: That the only point. `Intl` is a standard and the `dayjs` is not

Joozty:
I am not sure whether those icons shouldn't be smaller. At least they are in design.

viktor-yakubiv:
Yes, I made them smaller, they were too big.

Also, I was wrong about 'alt' attribute, we should find something better. The current `<title>` works as a `title` attribute.

I tried to decompose. However, I do not see data exposed from the API. Is there anything wrong?

Joozty:
Apparently all records match. I reported a similar issue. So hopefully it gets fixed soon.

PR #291, opened 2020-05-05
Show better description when all outputs are (non) compliant

Joozty:


pages/data-providers/[data-provider-id]/deposit-compliance.jsx
```diff
@@ -81,6 +81,31 @@ class PublicationDateColumn extends Table.Column {
...
+      value: 'All',
+      append: '',
+      caption: 'outputs are non-compliant',
+    }
+  }
+  return {
+    value: 'All',
+    append: '',
+    caption: 'outputs are compliant',
+  }
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-compliance.jsx:106]: We cannot say this. Our compliance level is an approximation. We could only say _We could not detect non-compliant outputs_ or something like that

texts/depositing/overview.yml
```diff
@@ -7,4 +7,10 @@ description:
   # Displays when we have deposit compliance calculated
   compliance-level: '{{amount}}% of outputs are non-compliant.'
 
+  # Displays when compliance level is 100%
+  all-compliant: All outputs are compliant.
```
viktor-yakubiv, line 11: The same here

pages/data-providers/[data-provider-id]/deposit-compliance.jsx
```diff
@@ -105,13 +130,7 @@ const DepositCompliance = ({
...
         <NumericValue
           tag="p"
-          value={valueOrDefault(
-            store.depositDates.complianceLevel &&
-              (100 - store.depositDates.complianceLevel).toFixed(2),
-            'Loading...'
-          )}
-          append="%"
-          caption="non-compliant"
+          {...getComplianceLevelNumberProps(store.depositDates.complianceLevel)}
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-compliance.jsx:133]: I do not think this should change. Maybe do not show `%` if the percentage is 0. It'd numeric value, so it's good to keep it numeric. I would not mentally expect to see _All_ instead the number.

texts/depositing/overview.yml
```diff
@@ -7,4 +7,10 @@ description:
   # Displays when we have deposit compliance calculated
   compliance-level: '{{amount}}% of outputs are non-compliant.'
 
+  # Displays when compliance level is 100%
+  all-compliant: We couldn't detect any non-compliant outputs.
```
viktor-yakubiv, line 11: I am not sure about informal language (_couldn't_) but let's keep it.

```suggestion
  all-compliant: We couldn't detect any non-compliant output.
```

PR #292, opened 2020-05-05
Improve loading message for cross repository check

Joozty:


viktor-yakubiv:
No, I am against it. It harms the user experience. It sounds like we are stupid enough to make this message because we are not smart enough to solve the problem.

We are going to sell this! This message is not sellable, at all! @oacore/api can we please spend some effort to speed the API up?

pages/data-providers/[data-provider-id]/deposit-compliance.jsx
```diff
@@ -173,9 +173,12 @@ const DepositCompliance = ({
                 store.depositDates.crossDepositLag.possibleBonusCount
               ),
             })
-          : 'Loading data'}
+          : 'Data is being processed. This may take several minutes.'}
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/deposit-compliance.jsx:176]: How do you dustinguish 'request is being processed' and 'we wait for the response'? API does not even send you 202 in such a situation. And if it does, we should do pooling or interval requests.

It's a not good path to go.

Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-compliance.jsx:176]: 'request is being processed' and 'we wait for the response' is the same and I don't see a reason need to distinguish them. It doesn't matter from user point of view and it makes it more complicated...

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/deposit-compliance.jsx:176]: 'Data is being processed' means you as a user send something and the system took it into account but it takes longer to process. When you did not send anything explicitly, did not press any button, it must not sound like that.

Joozty:
I am against saying we are stupid enough to make this message. I totally understand that something is really hard to get data fast considering man-hours and architecture. Like it's not ideal but it is what it is.

pages/data-providers/[data-provider-id]/deposit-compliance.jsx
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/deposit-compliance.jsx:176]: ok so it could have been changed to something like data is being fetched/calculated.

PR #297, opened 2020-05-06
Move URL control fully to Next Router

viktor-yakubiv:
This is a prerequisite for #285. Current dataProvider/activity controlling is too strict and would not work with anonymous users with no access to data providers. I propose to move control fully to Next.js and mirror router state to the store. If an error happens, handle it in route change error event.

@Joozty I am really sorry now that I pushed this, perhaps it was opinionated or unreasonable. It's my turn to change it now.


pages/index.js
```diff
@@ -1,19 +1,12 @@
...
   useEffect(() => {
-    const route = new Route({
-      dataProvider: store.dataProvider.id,
-      activity: store.activity.path,
-    })
-    router.push(route.href, route.as)
+    router.push('/data-providers')
   })
-  return null
 }
```
Joozty, line 10: you should probably leave `return null` here. Since router.push is async it doesn't mean that Home component gets redirected before render. It may happen after render.

viktor-yakubiv: I know. I hate that React cannot render `undefinded` and makes me to explicitly return `null`

Joozty: me too :(

viktor-yakubiv:
@Joozty it works cool, I am impressed even. However, I would be happy if you could help me with error handling providing advice or even pushing a commit.

viktor-yakubiv:
@Joozty this is ready for the review.

It does not work as good as I would like to, this 2-step-redirect is not good, but it works and this is a good step ahead.

I am prettifying my commits to make this able to merge. I feel like the code is not really good but I feel overwhelmed by changes scope to find what may be bad.

components/application/index.jsx
```diff
@@ -33,15 +32,17 @@ const Application = ({
...
-                dataProvider={dataProvider.id}
-                key={id}
-                value={id}
-                selected={id === activity.split('/')[0]}
-                path={path}
-              />
-            ))}
+            {activities.routes
+              .filter((activity) => activity.parent == null)
+              .map(({ path, test }) => (
```
viktor-yakubiv, line 37: This exposed activities configuration internals. I do not know how to do this better at the moment.

components/application/activities.js
```diff
@@ -0,0 +1,74 @@
...
+    return this.routes.find((route) => route.test(url))
+  }
+}
+
+const config = new RouteConfig([
+  {
+    test: /\/overview/,
+    path: 'overview',
+    icon: 'view-dashboard',
+  },
```
viktor-yakubiv, line 36: Doing so I come to the idea that we should use some library for route parsing or define our routes in a **much more** simple way.

components/application/index.jsx
```diff
@@ -33,15 +32,17 @@ const Application = ({
...
+                <ActivitySelect.Option
+                  dataProvider={dataProvider.id}
+                  key={path}
+                  value={path}
+                  selected={test(pathname)}
+                  path={path}
```
viktor-yakubiv, line 43: Maybe `value` or `path` is not needed anymore.

pages/_app/index.jsx
```diff
@@ -80,9 +77,11 @@ class App extends NextApp {
 
   fetchUser = async () => {
```
viktor-yakubiv, line 78: @Joozty why is this function needed?

pages/_app/index.jsx
```diff
@@ -143,11 +142,12 @@ class App extends NextApp {
     const { store } = this
     const { Component, pageProps } = this.props
     const { isAuthorized } = this.state
+    const pathname = this.props.router.asPath
```
viktor-yakubiv, line 145: Does not looks good, to be honest.

pages/data-providers/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+  </Card>
+)
+
+// It's not a real list. It's a redirect at the moment
+const DataProviders = ({ store }) => {
+  const { dataProviders } = store
+  const id = dataProviders && dataProviders[0].id
+
+  // TODO: Find a way to make this conditional
+  useRedirect('[data-provider-id]', `${id}`)
```
viktor-yakubiv, line 26: Maybe explicit `toString()` will let us know when something goes wrong, i.e. got `null` or `undefined`

```suggestion
  // TODO: Find a way to make this conditional
  useRedirect('[data-provider-id]', id.toString())
```

---
_TODO_ is still a question.

pages/_app/index.jsx
Joozty: It's among the first functions which are called when dashboard loads. It checks if a user is logged in or not. (Checking skipped when user navigates to pages which don't require log in.

viktor-yakubiv:
I often get _Cannot read property 'timeLagData' of null_. Probably because route is not reflected or Store does not have dataProvider selected at the request.

pages/_app/index.jsx
```diff
@@ -80,9 +77,11 @@ class App extends NextApp {
 
   fetchUser = async () => {
     try {
-      const { dataProvider, activity } = new Route(window.location.pathname)
-      await this.store.init(dataProvider, activity)
+      const { router } = this.props
+      await this.store.init(router.query['data-provider-id'])
```
Joozty, line 81: data-provider-id is undefined for some reason. Try to access some page directly http://core.dashboard:3000/data-providers/86/overview

Joozty: basically the same as https://github.com/oacore/dashboard/pull/297#issuecomment-624646926

pages/data-providers/[data-provider-id]/index.js
```diff
@@ -1,3 +1,8 @@
-const Index = () => null
+import { useRedirect } from '../../_app/hooks'
```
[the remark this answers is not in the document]
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/index.js:1]: It's tightly coupled to Next.js, I prefer to have it in `pages`. I would leave this as it at the moment. We can move it later if necessary.

pages/_app/hooks.js
```diff
@@ -0,0 +1,18 @@
...
+import { useRouter } from 'next/router'
+
+const useRedirect = (routePath, actualPath = routePath) => {
+  const router = useRouter()
+  const pathname = join(router.pathname, routePath)
+  const asPath = join(router.asPath, actualPath)
+
+  useEffect(() => {
+    router.push(pathname, asPath)
+  })
```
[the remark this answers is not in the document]
viktor-yakubiv: There is no reason in this. The second parameter is optional.

components/application/index.jsx
Joozty: I don't see a problem with it.

components/application/activities.js
Joozty: Routes will need to refactor if we want to show a warning icon in the menu near to issues.

pages/_app/index.jsx
Joozty: you can still destructure it in two lines above.

components/application/activities.js
viktor-yakubiv: I am considering inlining everything into JSX without this configuration.

I wonder if you have any other idea?

pages/_app/index.jsx
viktor-yakubiv: It's not about desctructuring. I did not want this for a reason, the purpose of this line is slightly different from the above ones.

I do not like passing pathname to application. It should be solved in another way but I do not know how exactly it should be.

pages/_app/hooks.js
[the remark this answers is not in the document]
Joozty: There is a reason because if it is called on every component render. It can make redirection several times. It's very unlikely but still.

components/application/index.jsx
Joozty: one can be removed I guess.

pages/_app/index.jsx
viktor-yakubiv: Should be resolved now. Could you please check one more time?

viktor-yakubiv:
Changing data provider still behaves weird

pages/_app/index.jsx
```diff
@@ -78,41 +80,39 @@ class App extends NextApp {
...
+      Sentry.configureScope((scope) => {
+        scope.setUser({
+          id: this.store.user.id,
+          email: this.store.user.email,
+        })
+      })
+
+      // Re-renders the component
+      // Initialises router.query with the current URL
+      this.softReload()
```
Joozty, line 112: So basically [nextjs keeps query empty in componentDidMount](https://nextjs.org/docs/routing/dynamic-routes#caveats) and fill it in next update. It's really annoying.


> Pages that are statically optimized by Automatic Static Optimization will be hydrated without their route parameters provided, i.e query will be an empty object ({}).
> 
> After hydration, Next.js will trigger an update to your application to provide the route parameters in the query object.

viktor-yakubiv: Thanks a lot! It's really helpful even if I do not know what to do with this now. I wonder how did you find this?

pages/_app/index.jsx
```diff
@@ -78,41 +80,39 @@ class App extends NextApp {
...
 
-    if (!this.isRouteWithoutStore) await this.fetchUser()
+    if (this.isRouteWithoutStore) return
 
-    // we use routeChangeStart to immediately update selected activity
-    // ensure it's bound after user is fetched to avoid false errors
-    //  - user has no access to data provider
-    router.events.on('routeChangeStart', this.handleRouteChange)
+    try {
+      await this.store.init()
```
Joozty, line 101: you need to pass here dataProviderID otherwise pages cannot be accessed directly. http://localhost:3000/data-providers/1/overview

pages/_app/index.jsx
Joozty: Well we can opt-out from Automatic Static Optimization for dynamic pages, i.e we need to add getInitialProps to _app.jsx or add getInitialProps to all pages in /data-provides/ folder. Eventually which is the best solution IMHO is to parse data provider id manually from windows.location in componentDidMount. 

I found it by browsing issues in next repo.

viktor-yakubiv: It seems to me that we should do the following:
1. keep `componentDidMount` sync
2. trigger store initialisation sync
3. subscribe for the `user` update (observe it)
4. once the `user` is changed, run route initialisation
   - at this stage, the router should be already dehydrated (🤔❓), `query` object is not empty

In the algorithm, I rely on the fact that there is an extremely small chance to have `user` initialised bebore router's `query` is. However, it's better to somehow observe the router. I wonder if the router triggers some events after initialisation, maybe `routeChangeComplete`.

pages/_app/index.jsx
viktor-yakubiv: It does work for me

pages/_app/index.jsx
Joozty: Sounds to me too complicated in comparison with the current state. (just parse provider from window object) Also, it's not good to rely on that fact even when there is a small chance... It can cause really hard debugging bugs.

pages/data-providers/[data-provider-id]/index.js
```diff
@@ -1,3 +1,8 @@
...
 
-export default Index
+const DataProvider = () => {
+  useRedirect('overview')
```
Comment (Joozty) [inline pages/data-providers/[data-provider-id]/index.js:4]: Maybe it's worth of redirecting on the server-side too? Just thinking out loud.

Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/index.js:4]: What do you mean? Add 302 response?

I would but I do not know how to do it.

pages/_app/index.jsx
viktor-yakubiv: It works because of `softReload`. It triggers route change and the store gets updated.

Joozty: Shoot sorry. I forgot to pull the latest changes. Nonetheless, it works just because the code is really fast and data provider is changed before next pretender, right?

pages/data-providers/[data-provider-id]/index.js
Comment (Joozty) [reply pages/data-providers/[data-provider-id]/index.js:4]: Something like this but I haven't tested it.

https://linguinecode.com/post/how-to-redirect-on-the-server-side-with-next-js

pages/_app/index.jsx
viktor-yakubiv: It's actually not. Next does work as I expected, `routeChangeComplete` fires after the client initialisation (dehydration, whatever). I subscribe to 2 promises and then run updates, as simple as it is; [check it out](https://github.com/oacore/dashboard/pull/297/commits/09cdb2ec4da419bbee1e0feba37b22e31a3def47).

pages/_app/index.jsx
viktor-yakubiv: Anyway, I changed it already

pages/_app/index.jsx
Joozty: Interesting. I though routeChangeStart and Complete are called always together. Is it just your assumption or did you find a proof for it? It may work like that just because routeChangeStart was fired before we managed to subscribe for it and since there is a delay between start and complete we manage to subscribe for the complete event. :)

pages/data-providers/index.jsx
viktor-yakubiv: @Joozty any ideas about conditional hooks? It will fail if an error happens and will redirect to `undefined`.

pages/_app/index.jsx
viktor-yakubiv: I ran a simple test with console logs marking each event. After initialisation only and always complete-one fires.

pages/data-providers/[data-provider-id]/index.js
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/index.js:4]: I created #300 for this.

pages/_app/index.jsx
viktor-yakubiv: I left `console.log` in my example. Feel free to check it, if everything is fine, I consider this PR ready

Joozty: Ok, you still haven't convinced me but let's see if it brings bugs. There are no much options anyway and this is the cleanest solution. :)

pages/data-providers/index.jsx
Joozty: Conditional hooks are bad practice there is even an eslint rule for it. I saw SWR has some sort of conditional hooks. Maybe you can have a look there or google something. Sorry, I am not much help here. It's new for me too.

viktor-yakubiv: I see the only possibility to pass there a function that either returns undefined if the hook is not needed or actual values if it is. 

Also, I can check first parameter to be truthy if hook is needed.

viktor-yakubiv:
It seems like I solved all comments. @Joozty could you please review it again? 🙂

Joozty:
Sorry, I wanted to do it sooner but I am having still troubles with API. Code looks good to me but when I access this page https://dashboard-git-redirects.oacore.now.sh/ it stays blank (just the app bar is there) and the user is fetched correctly with status code 200. Do you have the same problem?

viktor-yakubiv:
The same for me. It seems to be a bug, my guess it's because of `useEffect(..., [])` but I have to investigate it more.

viktor-yakubiv:
@Joozty this is a shitty workaround, I am really sorry about. This is the only thing I could find at the moment. I hope we will sort it out soon.

pages/_app/index.jsx
```diff
@@ -78,41 +76,70 @@ class App extends NextApp {
...
+    router.events.on('routeChangeComplete', logPageView)
 
-    window.addEventListener('unhandledrejection', this.handlePromiseRejection)
+    if (isRouteWithoutStore(router.asPath)) return
+
+    const { store } = this
+    const initializeUser = store.init()
+
+    /**
+     * TODO: Remove the following behaviour, find another option to handle this:
```
Joozty, line 95: I would rather suggest to not merging this. I mean I would prefer to push only the part with activities. I know it took you so much time but if we push this "dangerous code" it won't be touch. It will stay there forever. 😞 It's very similar to [this](https://github.com/oacore/dashboard/blob/master/pages/data-providers/%5Bdata-provider-id%5D/deposit-compliance.jsx#L140) 


I am not going to force you. It's your call here. 🙂

viktor-yakubiv: Honestly, I am not sure about the comment you reference to, it seems to be okay with me now. Also, this comment is going to be resolved in some way with templates like in #305, even if not, it should be removed.

About this, I wanted to merge this and refactor App straight after this PR. However, I am not sure we have a time frame for this refactoring so I will consider removing it since you have a strong point.

Thank you for the feedback! 🙂

viktor-yakubiv: I checked it one more time trying to parse the route during the initialisation to avoid dangerous code. I am afraid it's not possible and we have no other option at the moment.

The initial router state looks like this:
```js
{
  asPath: "/data-providers/[data-provider-id]/overview",
  pathname: "/data-providers/[data-provider-id]/overview",
  route: "/data-providers/[data-provider-id]/overview",
  query: {}
}
```

Not only `query` is not populated, but all parameters are not parsed.

Joozty: there is an option. parse `window.location` directly.

viktor-yakubiv: 🤦‍♂️

viktor-yakubiv: Does the current option satisfy you?

Joozty: yep, thank you! 🙂

PR #299, opened 2020-05-06
Redirect to the previous page after logging in

viktor-yakubiv:
For some reason, it behaves weirdly. I get an error from the API if 'continue' is passed to iframe!

### DIFF
--- a/pages/_app/index.jsx
+++ b/pages/_app/index.jsx
@@ -64,7 +64,9 @@ class App extends NextApp {
 
   redirectToLogin() {
     const { router } = this.props
-    router.push('/login')
+    const basePath = '/login'
+    const nextPath = router.asPath.indexOf(basePath) >= 0 ? '/' : router.asPath
+    router.push(`${basePath}?continue=${encodeURIComponent(nextPath)}`)
   }
 
   handlePromiseRejection = (event) => {
--- a/pages/login.jsx
+++ b/pages/login.jsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { withRouter } from 'next/router'
+import { useRouter } from 'next/router'
 
 import styles from './login.module.css'
 import Title from '../components/title'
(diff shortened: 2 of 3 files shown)


viktor-yakubiv:
@Joozty could you please review this?

The code seems to be the only correct one but I face 2 silly problems all the time described below.

### Unhandled promise rejection
I unexpectedly get unhandled promise rejection with 401 when I think that it should be handled in the Root Store

### Weird API redirect

When `continue` property is passed to `login.html`, after performing a login I get the following error in console:

> Access to fetch at 'https://api.dev.core.ac.uk/%2Fdata-providers%2F27%2Foverview' (redirected from 'https://api.dev.core.ac.uk/login_check') from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

I could not find any `continue` parameter sent to the API but this seems like API tries to detect a continue prop (maybe from the `Referer`) and redirect. I wonder if it really behaves like that but before checking this I would like to hear your qualification about.

Joozty:
> I unexpectedly get unhandled promise rejection with 401 when I think that it should be handled in the Root Store

If you mean next overlay warning which needs to be closed in order to continue then there is nothing to do about it I am afraid.[Nextjs subscribe for that event sooner](https://github.com/zeit/next.js/blob/canary/packages/react-dev-overlay/src/client.ts#L54) than us. We could create some sort of decorator for all our actions which would catch 401 errors and redirect to login or change stack frame to [something else than string](https://github.com/zeit/next.js/blob/canary/packages/react-dev-overlay/src/client.ts#L27).

Joozty:
Regarding your second point, I was unable to reproduce it. Tried to compare master with this branch and didn't see any difference. You sure about it? (tried localhost, 127.0.0.1 and core.dashboard)

viktor-yakubiv:
I have no idea why it happens. Let me try another browser, clear caches, cookies and other basic troubleshooting techniques. I will come back with a screenshot or recording.

Despite that, does it work as expected? Did you try logout and visit specific page, kind of `/data-providers/27/deposit-compliance` and then log in again?

viktor-yakubiv:
@Joozty I checked again in Firefox with cleared cache and cookies. It's the same. The last error in the console appears after pressing 'log in' button.

screenshot with error

viktor-yakubiv:
Ouch! 🤯

The API really checks `Referer` header for a continue parameter and redirects to that if it was present. @oacore/api could you please fix this?

PR #302, opened 2020-05-08
Run DepositDases requests sequentially

viktor-yakubiv:
@mcancellieri I am trying here to run requests sequentially. I am not sure it helps but this may try. I wonder if it's possible to test this without merging and deploying to most of our servers first.

@Joozty I hope I do it right 🙂

### DIFF
--- a/store/deposit-dates.js
+++ b/store/deposit-dates.js
@@ -31,9 +31,7 @@ class DepositDates extends Store {
     this.crossDepositLagCsvUrL = `${API_URL}${this.crossDepositLagUrL}?accept=text/csv`
     this.publicationDatesValidateUrl = `${baseUrl}/publication-dates-validate`
 
-    this.retrieveDepositTimeLag()
-    this.getPublicationDatesValidate()
-    this.getCrossDepositLag()
+    this.retrieve()
   }
 
   @computed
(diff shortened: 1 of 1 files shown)


store/deposit-dates.js
```diff
@@ -73,16 +71,24 @@ class DepositDates extends Store {
...
-  async getPublicationDatesValidate() {
+  async retrievePublicationDatesValidate() {
     const { data } = await this.request(this.publicationDatesValidateUrl)
     this.publicationDatesValidate = data
   }
+
+  async retrieve() {
+    // Doing it sequentially because of the API suffering
+    // from heavy concurrent requests
+    await this.retrieveDepositTimeLag()
```
Joozty, line 88: It should be executed differently otherwise if the first request fails the following will fail too. either wrap every request with try/catch or try to find a better solution. 

To be honest I am against doing these things because it looks to me like it's against javascript. This is something which should backend handle better. :(  If it is needed I can cope with it.

viktor-yakubiv:
```js
 await Promise.allSettled(retrieveDepositTimeLag(), ...)
```
Like this

Joozty: That doesn't guarantee sequential order.

viktor-yakubiv: Yeah, you are right.

Can be chained in
```js
this.retrieveDepositTimeLag().finally(...).finally(...)
```

if it's possible

viktor-yakubiv: Seems to work as expected

```js
let i = 0;

const exec = (timeout = 500) => 
  new Promise((resolve, reject) => {
    i++;
    console.log('call', i);
    setTimeout(() => {
      console.log('reject', i);
      reject(new Error(`Reject ${i}`));
    }, timeout);
  });

exec().finally(exec).finally(exec).catch(() => {});
```

PR #303, opened 2020-05-08
Show user name in AppBar when it's available

viktor-yakubiv:
I am still not sure I should do this but it looks better to me, at least because I have my name filled up.


Joozty:
would be nice if user could change their name in settings.

PR #305, opened 2020-05-11
Move DOI page to templates

Joozty:
@viktor-yakubiv this is how we could structure our page templates. Please have a look and let me know what do you think.


Joozty:
Let's connect only templates (not the cards) to store. Eventually, only pages can be connected. It's basically the same. I just want to avoid passing too many props from pages.

viktor-yakubiv:
I see 2 options at the moment:
1. connect `pages` and pass props down
2. connect `modules` and do not pass props at all

We do not have `modules` yet. I imagine this kind of directory where we have a module connected to a single time of store, e.g. `deposit-compliance`, `user`, `content`/`works` and so on.

Maybe the first option fits us better at the moment. It's less work.

PR #308, opened 2020-05-12
Move Content to templates

Joozty:


pages/_app/index.jsx
```diff
@@ -14,6 +17,9 @@ import { initStore, GlobalProvider } from 'store'
 import Application from 'components/application'
 import { Sentry } from 'utils/sentry'
 
+dayjs.extend(relativeTime)
+dayjs.extend(advancedFormat)
```
viktor-yakubiv, line 21: What do these plugins do?

Can we please get rid out from `dayjs` in favour of `Intl.DateFormat`?

Joozty: Once issues are merged we can.

PR #309, opened 2020-05-12
Fix now.sh deployment

Joozty:


Joozty:
It's should be quite easy to resolve conflict as it is only renaming the files. I can handle that no problem. :)

viktor-yakubiv:
If you marked this as a draft because you want to wait till #287 merge, let's merge this one and sort the issues out later. I am not sure, I will finish it soon and do not want to block deploymnets, it's frustrating to work like that.

Joozty:
I marked it as a draft to disable merge functionality. I will keep this branch up to date with master a few days. If now.sh doesn't fix it till then we can merge it...

PR #312, opened 2020-05-12
Handle 202 Accepted status code

Joozty:


Joozty:
@viktor-yakubiv there is no way to try it on real example yet. You may check code if it looks good to you. I mocked it and it worked just fine. :)

Joozty:
I had to move request back to constructor because the loading indicator got broken by that change.

Joozty:
Maybe it would work if `defautltOptions` is not static and request use `.bind`. Anyway, it's better to use it like this because now the loading indicator shows only on the first request.

viktor-yakubiv:
I prefer to have fewer changes and clear diffs for each commit. Sorry, I changed that.

I prefer binding in this case because the code seems to be more natural (check the first commit, I got it from #336).

store/root.js
```diff
@@ -12,14 +12,27 @@ import { AccessError, AuthorizationError, PaymentRequiredError } from './errors'
...
 import * as NetworkErrors from 'api/errors'
 
+const REPEADTED_REQUEST_TIMEOUT = 30000
+
 class Root extends Store {
   static defaultOptions = {
     allowAnonymousAccess: false,
 
     request(url, options) {
+      const requestContinously = () =>
```
Joozty, line 22:
```suggestion
      const requestContinuously = () =>
```

store/root.js
```diff
@@ -12,14 +12,27 @@ import { AccessError, AuthorizationError, PaymentRequiredError } from './errors'
...
+        apiRequest(url, options).then((response) =>
+          response.status === 202
+            ? new Promise((resolve, reject) => {
+                const repeadedRequest = () =>
```
Joozty, line 26:
```suggestion
                const repeatedRequest = () =>
```

store/store.js
```diff
@@ -11,7 +11,7 @@ class Store {
   }
 
   request(...args) {
-    return this.options.request(...args)
+    return this.options.request.call(this, ...args)
```
Joozty, line 14: Sadly this doesn't cover table case. I am not sure why...

viktor-yakubiv: Yes, I was thinking about tonight and I actually found the problem. This code binds to the current `this`, i.e. to each individual `Store` instance. Despite, we need to bind it to the certain `Root` instance.

I am adding a fix for this.

viktor-yakubiv:
@Joozty can I merge this? It has conflicting changes with #336 which I would like to solve.

Joozty:
It doesn't work even after fixing typo. :(

store/root.js
```diff
@@ -44,6 +44,13 @@ class Root extends Store {
     },
   }
 
+  constructor(options) {
+    super(null, {
+      ...options,
+      reqeust: (args) => Root.defaultOptions.request(...args),
```
Joozty, line 63:
```suggestion
      request: (args) => Root.defaultOptions.request(...args),
```

viktor-yakubiv: Because there is another typo:

```suggestion
      request: (...args) => Root.defaultOptions.request(...args),
```

Joozty: Good catch but sadly it's not enough.

viktor-yakubiv: If it's not, I will bind the function. This will solve the problem

viktor-yakubiv:
I am not satisfied with the solution but there is no better option at the moment. 

Our API layer started causing problems. Error processing duplicates in Store and API, we should stop treating status code ≥400 being an error.

PR #314, opened 2020-05-12
Hide activities when user has been accidentally logged out

Joozty:
@viktor-yakubiv Please bear in mind that this has to be handled in https://github.com/oacore/dashboard/pull/297 too. :)

### DIFF
--- a/pages/_app/index.jsx
+++ b/pages/_app/index.jsx
@@ -78,6 +78,9 @@ class App extends NextApp {
 
   handlePromiseRejection = (event) => {
     if (event.reason instanceof UnauthorizedError) {
+      this.setState({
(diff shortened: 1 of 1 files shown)


viktor-yakubiv:
I am not so sure about this. I would prefer to cache the store continuously and subscribe for the cache updates. Or add a message broadcasting for the store to sync changes between tabs. When the tab is focused, store invalidates. [Read the thread](https://stackoverflow.com/questions/28230845/communication-between-tabs-or-windows) to find out more about.

Although, I can add this to #297 for the time being.

Joozty:
I didn't try to cover multiple tabs. This cover only the case when API logs you out for any reason...

PR #315, opened 2020-05-13
Use correct title for all pages

Joozty:


components/title/index.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import Head from 'next/head'
+import { classNames } from '@oacore/design/lib/utils'
+
+const Title = ({ children, hidden = false, tag: Tag = 'h1' }) => {
+  if (!(typeof children === 'string'))
+    throw new Error('Unsupported children for Title component')
```
viktor-yakubiv, line 7: Throw this only in `development`. There is no reason to unexpectedly unmount full tree if the title is not correct. If I understand, if there is HTML, it will be dropped in the `title` element. so there is no problem with that.

PR #317, opened 2020-05-13
Add IRUS card

Joozty:


Joozty:
@viktor-yakubiv Can you please instruct me what's should be shown in this card? Is the chart vertically centred enough? Do we want to show some description or chart tooltip? Thanks

templates/overview/cards/irus-card.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+
+import OverviewCard from './overview-card'
+
+import { Card } from 'design'
+import StatisticsChart from 'components/statistics-chart'
+
+const IrusCard = () => (
+  <OverviewCard>
+    <Card.Title tag="h2">IRUS statistics</Card.Title>
```
viktor-yakubiv, line 10: IRUS-UK = Institutional Repository Usage Statistics UK

_Statistics_ is duplicated.

Joozty:
I think we should rather set min-width to the card.

viktor-yakubiv:
Yeah, we are doing much work today 😄 🎉

Joozty:
I have a feeling that they decreased the number of free deployments per day

viktor-yakubiv:
Yeah, seems to be so. It's 100 per day now.

viktor-yakubiv:
I rebased to force redeploy. Also, I used more realistic data to display the chart.

Joozty:
@viktor-yakubiv The irusUKUrl hasn't been implemented yet. If you don't mind I wouldn't postpone this PR. We can add it later.

viktor-yakubiv:
@mcancellieri and @PMRuma this seems to be finished. I wonder if you have time to review it or just check how it looks.

PR #321, opened 2020-05-13
Issues review

viktor-yakubiv:
I moved issues from the [document](https://docs.google.com/document/d/19si9SjAo61VIj7Em4v2LjzDfP3HZgeB2FHP62oHYHtM/edit#) into the YAML file we are using to display issues explanation and resolution in the Dashboard.

@oacore/operations, especially @samuelpearce, we will need to carefully review this all together, finalise it and populate the outputs to all necessary places.

@oacore/api below you may found a few items that are marked `hidden: true`. This means that the issue is (will be) reported by the system but **should not be displayed** in the Dashboard. We have to find a way to filter it out ideally on the API side. This filtering impact counters so we will need to either find a workaround on the front-end or find a way to filter it out on the back-end for the Dashboard but maybe disable filtering with direct request, what seems to be workaround too. Although, @samuelpearce could you please explain how you imagine this?

## File structure

You may find the data structure explained in the document header. The file consists of sections described below.

The first one (_OLD ONES_) lists the issues that are exported by the system currently or at least exist in the _Issue_ enumeration but I could not find any corresponding issue in the document to merge it.

From the second section till the end of the file are new issues with their raw explanation that **should be properly edited**. The `type` attribute is the new enumeration name of the issue. If the exists `matches` attribute, it points to the old name of the issue if there is any.

You can find renaming in the git diffs of the first commit in this PR (check _Changes_ tab). To simplify identification locally I populated **empty `matches` to the new issues** that do not have corresponding ones in the current system structure and **omitted `matches` for already existing issues** that have not been renamed.

### DIFF
--- a/texts/issues/messages.yml
+++ b/texts/issues/messages.yml
@@ -1,63 +1,453 @@
--
-  type: OAI_ENDPOINT
-  title: There were problems with an OAI endpoint
--
-  type: INVALID_OAIPMH_ENDPOINT
-  title: Invalid OAI-PMH endpoint
+#!
+# Guidelines
+# ==============================================================================
+#
+# Data structure
+# ------------------------------------------------------------------------------
+#
+# The following properties are required and widely used in the project
+#
+# - `type`          unique issue identifier, user to identify the messages
+#
+# - `matches`       list of types that issue may appear with,
+#                   needed to support past and future renaming
+#
+# - `title`         short description in a few words
+#
+# - `description`   detailed description of the issue
+#
+# - `resolution`    instructions to resolve the issue
+#
+# - `severity`      the level of the issue impact
+#
+#
+# The rest is only informative, needed for this document self-descriptiveness:
+#
+# - `trigger`       the reason why the issue happens
+#
+# - `details`       what is included into the details object,
+#                   on overview what variables could appear in them `message`
+#                   and `resolution` properties
+
+
+# OLD ONES
+# ==============================================================================
+#
+# We have to merge this ones with the bottom list
+
 -
   type: SSL_CERTIFICATE_ERROR
   title: SSL certificate error
--
-  type: GENERIC_OAI
-  title: Generic OAI
--
-  type: GENERIC_ATTACHMENT_URL
-  title: Generic document URL
--
-  type: ROBOTS
-  title: Problems with the configured robots.txt
--
-  type: ATTACHMENT_MALFORMED_URL
-  title: A document URL is malformed
--
-  type: ATTACHMENT_NOT_VALID
-  title: A document is not valid
--
-  type: NO_FULL_TEXT_LINKS
-  title: No attachments found
+  severity: DEBUG
+  level: internal
 -
   type: EXTERNAL_UNKNOWN
-  title: No full text was found due to unknown external reasons
+  # title: No full text was found due to unknown external reasons
+  severity: DEBUG
+  level: internal
+  description: |
+    We were unable to identify what kind of error happend during the harvesting.
+    If this issue impacts a lot of records, please contact us.
+
+
+# Metadata Download
+# ==============================================================================
+#
+# Mostly affects the whole repository.
+
 -
-  type: ENCRYPTED_ATTACHMENT
-  title: Full text has digital restrictions
+  type: RESUMPTION_TOKEN_FAILURE
+  matches: [OAI_ENDPOINT]
+  severity: ERROR
+  trigger: |
+    When we are unable to download due to a failed resumptionToken
+  title: OAI-PMH resumption token failure
+  description: |
+    The resumption token issued results in a failed request
+
+    {{ url }}
+  resolution: ''
+  details: [url]
+
+-
+  type: INVALID_OAIPMH_ENDPOINT
+  severity: ERROR
+  trigger: |
+    If the OAI-PMH harvest fails, and it is caused because of an incorrect
+    OAI-PMH endpoint
+  title: OAI-PMH endpoint is not valid
+  description: |
+    Your current OAI-PMH endpoint ({{ oaiPmhEndpoint }}) is not valid.
+    If you changed the endpoint recently, please, update it in
+    [Settings](./settings#oai-pmh).
+  resolution:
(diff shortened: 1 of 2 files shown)


texts/issues/messages.yml
```diff
@@ -1,63 +1,444 @@
...
+
 -
   type: ATTACHMENT_NOT_VALID
-  title: A document is not valid
+  severity: ERROR
+  trigger: |
+    We check that a download is a pdf by checking the first 5 bytes contain
+    %pdf- If this is false, we set the PDF_NOT_VALID. In most cases, the issue
+    is replaced by a more specific issue as this issue is always triggered
+    when downloading HTML files.
```
viktor-yakubiv, line 189: @samuelpearce what kind of HTML files we download?

texts/issues/messages.yml
```diff
@@ -1,63 +1,444 @@
...
+  trigger: |
+    After downloading an HTML page, we attempt to extract links. If 0 are
+    found, this issue is triggered.
+  title: No valid attachment links found
+  description: |
+    We downloaded the repository metadata page for this item, but we were
+    unable to find any full text links
+  resolution: |
+    Ensure that dc:identifier contains a direct link to the full text. We only
+    support the following file types: .pdf
```
viktor-yakubiv, line 213:
```suggestion
  description: |
    We downloaded the metadata, but we were unable to find any full text links.
    Ensure `dc:identifier` contains a direct link to the full text. We only
    support `.pdf` file type.
  resolution:
    -
      caption: Mark resolved
```

texts/issues/messages.yml
```diff
@@ -1,63 +1,444 @@
...
+    viewing the full text via a screen reader.
+
+    If this is message is issued with a Warning, we were able to bypass the
+    security settings
+
+    If this message is an error, we were unable to bypass the security and we
+    have not imported the fulltext.
+  resolution: |
+    We recommend that Document Security is removed from the PDF.
+    [https://helpx.adobe.com/uk/acrobat/using/securing-pdfs-passwords.html#remove_password_security]
```
viktor-yakubiv, line 256:
```suggestion
  trigger: |
    This issue is triggered when a pdf has DRM enabled.

    If we are able to bypass, a WARNING is given, if we are not, an ERROR.
  title: The full text is encrypted
  description: |
    The PDF file has Document Security enabled. This digitally restricts how the
    document is opened and used. For example, it may prevent editing, printing
    or viewing the full text via a screen reader.

    We recommend to remove Document Security from the full text document.
  resolution:
    -
      caption: Learn more
      url: https://helpx.adobe.com/uk/acrobat/using/securing-pdfs-passwords.html#remove_password_security
    -
      caption: Mark resolved
```

texts/issues/messages.yml
```diff
@@ -1,63 +1,444 @@
...
+    dc:title
+    A study of Open Access Papers in Europe
+
+    Studying Open Access papers in Europe
+
+    The closer the match, the more likely we will accept the change.
+
+
+    You may bypass this check by including a link to the full text in
+    dc:identifier
```
viktor-yakubiv, line 299:
```suggestion
  title: Full text title mismatches metadata
  description: |
    To ensure we link the correct metadata and full text, we check that the
    metadata title text is contained within the PDF document.
    An exact match is not required, but must be similar.

    This error is common for scanned items where Optical Character
    Recognition&nbsp;(OCR) has not been performed.

    Ensure that the Metadata Title is the same as the document title
    or including a link to the full text in `dc:identifier`.
  resolution:
    -
      caption: Mark resolved
  # should be moved to separate place or deleted
  extra-intructions: |
    Ensure that the Metadata Title is the same as the document title. For
    example:

    Failed match example:
    - `dc:title:` _An Open Access Paper: a study_
    - PDF title: _A study of Open Access Papers_

    A successful partial match:
    - `dc:title`: _A study of Open Access Papers in Europe_
    - PDF title: _Studying Open Access papers in Europe_

    The closer the match, the more likely we will accept the change.

    You may bypass this check by including a link to the full text
    in `dc:identifier`.
```

texts/issues/messages.yml
samuelpearce: Usually when metadata provides a link to an HTML page such as [this page](https://aura.abdn.ac.uk/handle/2164/14314).

When we extract links from the HTML page and crawl for them, we check the downloaded file that it is a valid and well-formed pdf

texts/issues/messages.yml
samuelpearce: You've changed the title from PDF to full text, but the description is very specific for pdfs 

If we do end up supporting other documents, we'll either need a way to report different messages for each file format, or each file format requires it's own issue status for encrypted content. Any thoughts on this?

samuelpearce: Re: learn more, it is possible to have multiple "learn more" links?

viktor-yakubiv: It's possible to have as many links as we want. However, from the point of user experience, I prefer to have no more than 2.

I agree with the comment about the title and will change it back. Thank you!

texts/issues/messages.yml
samuelpearce: Line 269: PDF documents to full text
Line 274: is `resolution` missing from this line?

texts/issues/messages.yml
viktor-yakubiv: However, I am not so sure. I wanted to link the issue to the part of the metadata right from the title. We expose this part (PDF) as a 'Full text' property. I wanted to point out that the problem happened there.

Also, from the point of the system, I am not sure we will download and expose multiple documents in different formats even if we support multiple formats.

texts/issues/messages.yml
viktor-yakubiv: Line 269: 👍 changed

Line 274: No, it's not. The idea is to have a single description and under the resolution have a set of action buttons that either go to some kind of settings, instructions or dismiss the issue.

viktor-yakubiv:
I mark this ready to review but this will be automatically merged via #500.

PR #322, opened 2020-05-13
Upgrade packages

viktor-yakubiv:
Blind execution of `npm upgrade`. This still needs to be reviewed.

I am trying to get rid out from console warnings about `caniuse-lite` and `unfetch.js`


viktor-yakubiv:
`package.json` looks okay to me. @Joozty what do you think?

viktor-yakubiv:
Also, the project builds and runs. I wonder if I need to check anything else 😄

viktor-yakubiv:
Wow! This looks nice 😀

new error messages

PR #334, opened 2020-05-16
Try Nivo

viktor-yakubiv:
@Joozty I've been exploring [awesome-charting](https://github.com/zingchart/awesome-charting) yesterday and found [a new gem](https://github.com/zingchart/awesome-charting/pull/29) that is really awesome!

I was really excited to try [Nivo](https://nivo.rocks) in our project. The library supports both SVG and Canvas! Also, it ships each chart type in a separate package and has the API designed exactly in a way I wanted. This's never happened before!

Personally, I am bought to move all charts to the library. I find the bar chart much easier to use, and the pie chart covers the case we have a workaround now.

I am looking forward to hearing your feedback about.

### DIFF
--- a/components/time-lag-chart/index.jsx
+++ b/components/time-lag-chart/index.jsx
@@ -1,20 +1,44 @@
 import React from 'react'
+import { ResponsiveBar as BarChart } from '@nivo/bar'
 
-import { toPoints, withGaps, withBuckets, toBucketSum } from './data'
-import TimeLagChart from './chart'
+// WIP: Assuming that we will or already do use aggregation in the target PR
+/* eslint-disable no-unused-vars */
+
+import {
+  toPoints,
+  toX,
+  toY,
+  withGaps,
+  withBuckets,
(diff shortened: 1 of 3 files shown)


Joozty:
If you can guarantee a smooth transition between these two chart libraries and are willing to rewrite them then why not. My only worry is that we are finalising this project and trying to move to search and we might end up with two chart libraries here. 

Having every chart in extra package is a big plus. They have also nicer docs. On the other hand, I am not that excited about the API. I don't see a big difference based on your example here but I only had a very brief look.

viktor-yakubiv:
I am sure it will be better and the transition will not be a problem. I will handle it. However, will start when #268 is finished.

I am converting this PR to a draft now to prevent merging and will adjust it to the most recent `master` state when needed.

Thanks a lot for your feedback! You may not see it from my example. To me, it looks really promising because I removed a file doing this change but I did not do it, only stopped using the file that draws actual chart.

PR #335, opened 2020-05-18
Payment required error population

viktor-yakubiv:
Working on store interface for 402 error handling

### DIFF
--- a/store/helpers/pages/pages.js
+++ b/store/helpers/pages/pages.js
@@ -3,12 +3,15 @@ import { observable } from 'mobx'
 import getOrder from '../order'
 import invalidatePreviousRequests from '../invalidatePreviousRequests'
 import Store from '../../store'
+import { PaymentRequiredError } from '../../errors'
 
 import { NotFoundError } from 'api/errors'
 
 const PAGE_SIZE = 100
 
 class Pages extends Store {
+  @observable error = null
+
   @observable data = []
 
   searchTerm = ''
@@ -59,8 +62,8 @@ class Pages extends Store {
 
     const request = this.request(this.url, { searchParams: params, signal })
     return new Promise((resolve, reject) =>
-      request.then(
-        ({ data, headers }) => {
+      request
+        .then(({ data, headers }) => {
           if (this.pageNumber === 0) {
             this.isFirstPageLoaded = true
             const length = headers.get('Collection-Length')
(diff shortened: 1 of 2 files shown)


store/root.js
```diff
@@ -14,47 +14,45 @@ import apiRequest from 'api'
...
 
 class Root extends Store {
-  options = {
+  static defaultOptions = {
     allowAnonymousAccess: false,
-  }
 
-  constructor() {
-    const request = async (url, options) => {
+    request(url, options) {
```
Joozty, line 20: @viktor-yakubiv Can you create from this commit PR. It's a bug. I solved this issue slightly differently and I like your solution more. Also, we will avoid conflicts.

viktor-yakubiv:
Contains 2 commits from #339

store/helpers/pages/pages.js
```diff
@@ -77,16 +80,25 @@ class Pages extends Store {
...
+          this.isFirstPageLoaded = true
+          this.isLastPageLoaded = true
+
+          if (error instanceof NotFoundError) {
+            this.data = []
+            resolve()
+          } else if (error instanceof PaymentRequiredError) {
+            // Stripping data to 5 elements only if there are more
+            // These elements may be used as a preview
+            this.data = this.data.slice(0, 5)
```
Joozty, line 98: Btw. what if data provider doesn't have any data? Should we still show the placeholder? We are basically selling no data. My point is whether 402 take precedence over 404. I would rather suggest to have it the other way around if it's not that way already. It's more a question for @oacore/api.

viktor-yakubiv: I added a trick around this into the store. I agree, 404 should be more important. @oacore/api plase consider the following error precedence for _protected_ endpoints:
1. `401 Unauthorized` if the user is anonymous **(highest priority)**
2. `403 Forbidden` if the user is logged in but trying to get a resource they do not have access to
3. `404 Not Found` if the user has access rights but there is no data to provide (empty object or collection is returned)
4. `402 Payment Required` if the collection is not empty but the user is a free customer and cannot get access to the resource **(lowest priority)**

viktor-yakubiv:
Closing this in favour of #336. It's hard to track it separately and update accordingly.

store/helpers/pages/pages.js
mcancellieri: Using this we need to run the full endpoint even when it shouldn't be allowed.
I would argue that 402 should be returned any case with data or not, therefore I would swap (3) and (4). 
I like the connection of status codes with semantic but I am not completely bought on using 404 so strictly for all the empty results.

viktor-yakubiv: > It could be that there is no data in that endpoint but we are still selling a valuable product and data from other endpoints!

Exactly! I do not want to display the message in the wrong place. If something is not sellable we must not annoy the user.

It's done with a workaround now because I do not even know the real precedence. I would like to avoid this trick but feel free to swap (3) and (4) if you find it painful or unreasonable to do it the proposed way.

PR #326, opened 2020-05-14
Rewrite table to functional component

Joozty:

PR #336, opened 2020-05-18
Form for requesting premium version

viktor-yakubiv:
~Working on~ The form for requesting premium. Read commit messages for the detailed description.

~`402` as a base used for testing only. This must be merged to `master` first and #335 aligned after.~

### DIFF
--- a/components/forms/request-premium/index.jsx
+++ b/components/forms/request-premium/index.jsx
@@ -1,0 +1,98 @@
+import React, { useState, useCallback } from 'react'
+import { classNames } from '@oacore/design/lib/utils'
+
+import styles from './styles.module.css'
+
+import { Button, TextField } from 'design'
+
+// TODO: Should be moved to @oacore/design
+const Textarea = ({
+  id,
+  className,
+  label,
+  tag: Tag = 'p',
+  ...textareaProps
+}) => (
+  <Tag id={id} className={classNames.use(styles.textarea).join(className)}>
+    <textarea id={`${id}-textarea`} {...textareaProps} />
+    <label htmlFor={`${id}-textarea`}>{label}</label>
+  </Tag>
+)
+
+const Form = ({
+  // data props
+  fromEmail,
+  fromName,
+  subject,
+  body,
+  product,
+
+  // html props
+  children,
+  className,
+  ...formProps
+}) => {
+  // Showing name only when email is edited
+  const [showName, toggleName] = useState(false)
+
+  // Showing message only when user decided to edit it
+  const [showMessage, toggleMessage] = useState(false)
+
+  const handleMessageToggle = useCallback(() => toggleMessage(true), [])
+
+  return (
+    <form
+      className={classNames.use(styles.container).join(className)}
+      {...formProps}
+    >
+      <div className={styles.personal}>
+        <TextField
+          type="email"
+          name="email"
+          defaultValue={fromEmail}
+          autoComplete="email"
+          label="Email"
+          helper="We will get back to you using the above email address"
+          onChange={useCallback(() => toggleName(true), [])}
+          tag="p"
+        />
+        {showName ? (
+          <TextField
+            name="name"
+            defaultValue={fromName}
+            autoComplete="name"
+            label="Name"
+            tag="p"
+          />
+        ) : (
+          <p>
+            <input type="hidden" name="name" value={fromName} />
+          </p>
+        )}
+      </div>
+
+      <input type="hidden" name="subject" value={subject} />
+
+      {showMessage ? (
+        <Textarea id="body" name="body" label="Message" defaultValue={body} />
+      ) : (
+        <p>
+          <input type="hidden" name="body" defaultValue={body} />
+          <Button
+            variant="outlined"
+            type="button"
+            onClick={handleMessageToggle}
+          >
+            Customise message
+          </Button>
+        </p>
+      )}
+
+      <input type="hidden" name="product" value={product} />
+
+      {children}
+    </form>
+  )
+}
+
+export default Form
--- a/modules/billing/payment-required/data.js
+++ b/modules/billing/payment-required/data.js
@@ -1,0 +1,108 @@
+import { useEffect } from 'react'
+import { autorun, observable } from 'mobx'
+import { useObserver } from 'mobx-react-lite'
+
+import { requestPremium as letter } from 'texts/letters'
+
+const globalContext = observable({
+  user: {},
+  organisation: {},
+})
+
+const userInput = observable({
+  update(patch) {
+    Object.assign(this, patch)
+  },
+})
+
+const emailContext = observable({
+  toEmail: 'enterp\u0072is\u0065\u0040cor\u0065\u002e\u0061\u0063\u002euk',
+
+  // These 4 are mirrored from the global store
+  // with and React's effect and MobX's autorun in useSync()
+  organisationId: '',
+  organisationName: '',
+
+  get fromName() {
+    return userInput.name ?? globalContext.user?.name
+  },
+
+  get fromEmail() {
+    return userInput.email ?? globalContext.user?.email
+  },
+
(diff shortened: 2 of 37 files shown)


viktor-yakubiv:
@Joozty could you please do some early review? It seems to be okay but needs to be prettified and properly integrated.

viktor-yakubiv:
@Joozty I consider this finished and ready for the review. The first commit fixes a bug introduced by #337. 

Although, I have to test UX of the form itself and consider @mcancellieri's remark that it's quite long. Also, I need to test actual form sending and see how it works, I missed a few hidden inputs and additional values to make it work.

modules/billing/payment-required/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+      )
+    },
+    [send]
+  )
+
+  // Populate changes on blur only to prevent live editing in two places
+  const handleChange = useCallback((event) => {
+    const { name, value } = event.target
+    data.update({ [name]: value })
+  })
```
Joozty, line 45:
```suggestion
  }, [])
```

modules/billing/payment-required/data.js
```diff
@@ -0,0 +1,103 @@
...
+
+  update(...args) {
+    userInput.update(...args)
+  },
+})
+
+const useSync = (store) =>
+  useEffect(
+    () =>
+      autorun(() => {
```
Joozty, line 68: I see now why you wanted to export store. You made it real reactive. 👍 To me both store or this autorun is good. Maybe I would just correct my statement and say that exporting store is better because it doesn't make duplicate values even they are kept in sync. (Even if testing or whatsoever may be harder).

I am quite surprised why you decided to go this way.  Not saying it's a bad decision. I imagined using only `withGlobalStore` HOC and nothing else in modules. Importing mobx stuff here is making more files dependent on it.

modules/billing/payment-required/data.js
```diff
@@ -0,0 +1,103 @@
...
+})
+
+const userInput = observable({
+  update(patch) {
+    Object.assign(this, patch)
+  },
+})
+
+const emailContext = observable({
+  toEmail: 'enterp\u0072is\u0065\u0040cor\u0065\u002e\u0061\u0063\u002euk',
```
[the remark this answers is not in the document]
viktor-yakubiv: To prevent SPAM scanners detect and collect email from the JS code or from source code (that is why it's here obfuscated too). I am not sure it's needed, @samuelpearce used to do so, I think it's not bad to protect.

Despite it leaks to the HTML in the source look, I do not mind it in the closed project. However, we may add some protection there too.

modules/billing/payment-required/data.js
```diff
@@ -0,0 +1,103 @@
...
+        // and then changed own name, it would not be confusing because
+        // it should be expected.
+        globalContext.user = store.user
+        globalContext.organisation = store.organisation
+
+        emailContext.actionUrl = store.contactUrl
+        emailContext.send = store.sendContactRequest
+      }),
+    [store]
+  )
```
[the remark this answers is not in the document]
viktor-yakubiv: I do not think disposer needs some context binding, so it can be returned directly. Do you want to change it because of readability?

modules/billing/payment-required/data.js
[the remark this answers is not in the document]
Joozty: Interesting, that's quite clever.

modules/billing/payment-required/data.js
viktor-yakubiv: I wend this way because if the `user`, for example, is set to `null`, then to an object, then the object, we will not track it. Although, if we use `autorun` we will do.

This option looked kind of interesting to me, kind of fun. It does not have to be good, I do not really know what is better.

---

> I imagined using only `withGlobalStore` HOC and nothing else in modules. Importing mobx stuff here is making more files dependent on it.

I wanted to have some kind of shared local state. We repeat the form in a few places, so I wanted to have changes populated everywhere if the user edited a single one.

It saw 2 options for this: React Context and MobX. I went to MobX because it seemed to be easier, `useObserver` hook would automatically run updates for me. Although, uncontrolled components slightly prevent this idea. However, if use controlled inputs, we will need a more sophisticated solution.

components/forms/request-premium/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+    >
+      <div className={styles.personal}>
+        <TextField
+          type="email"
+          name="email"
+          defaultValue={fromEmail}
+          autoComplete="email"
+          label="Email"
+          helper="We will reach you back using this email address"
+          onChange={useCallback(() => toggleName(true), [])}
```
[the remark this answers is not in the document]
viktor-yakubiv: I thought it should have an effect here but I am not sure. 

Working with hooks closer I find it quite interesting and tricky. Their behaviour is not obvious even if the interface is quite clear. I assume that hook call is replaced with something during compilation to prevent, for example, creating a function on every render in this case.

I should learn it more deeply to be sure in what I say.

modules/billing/payment-required/data.js
Joozty: Using MobX was a good way to go. I would personally create a new sub store for it not mixing state and UI component together. You taught me that. 🙂

components/forms/request-premium/index.jsx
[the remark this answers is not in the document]
Joozty: You are right. I tested it and it has an effect. :)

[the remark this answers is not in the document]
viktor-yakubiv: How did you test it?

templates/doi/cards/table-card.jsx
```diff
@@ -24,7 +26,12 @@ const formatDOI = (entity) => {
 const TableCard = ({ pages }) => (
   <Card className={styles.doiTableCard} tag="section">
     <Card.Title tag="h2">Browse DOI records</Card.Title>
-    <Table pages={pages} className={styles.doiTable} searchable>
+    <Table
+      pages={pages}
+      className={styles.doiTable}
+      excludeFooter={pages.data.length === 0}
```
Joozty, line 32: This is a little bit tricky. Data are "temporarily" nulled when user try to search. So it may cause a flickering effect.

In https://github.com/oacore/dashboard/pull/326 I removed `isFirstPage` flag and determine it out of data. If data is null it means data is loading if data === [] it means first page is loaded.

viktor-yakubiv: I am not sure, I wanted to rely on data absence. If data is emptied on search, it's fine, if it's nulled — it should be emptied.

The problem is that we should disable search and sorting if there is an error.

Joozty: Disabling search should be a problem. Sorting may be a bit problematic. I think a new functional table supports change in children props. I need to check it more. Class table computed sorting configuration in the constructor. Another possibility is to disable sorting globally by some flag.

viktor-yakubiv: As you said, disabling search was not a problem. I did not manage to work out with sorting so I will wait for #326.

@oacore/api we will need to disable sorting and filtering on backend to achieve data consistency in the 402 case.

viktor-yakubiv:
@mcancellieri I wonder if you have time to review this. If you do not I have no other way than merge it as is. Go through the commits, most of them are just preparation for the feature that is integrated by the last one.

@oacore/front-end I still experience problems. The request-premium form appears only after page reload or on the second visit, i.e. when the deposit dates page was opened once, no error appears but on the second step it's shown.

viktor-yakubiv:
> The request-premium form appears only after page reload or on the second visit, i.e. when the deposit dates page was opened once, no error appears but on the second step it's shown.

The bug is sorted.

There is another minor bug when you click on the row. The white sidebar appears, and it's not possible to close it. I would keep it for now and handle it later.

PR #338, opened 2020-05-20
Remove log timing

Joozty:


Joozty:
It's unused and I don't like having unused code in repo. if we need it we can still add it back.

PR #339, opened 2020-05-20
Add '202 Accepted' handling for cross-deposit-lag

viktor-yakubiv:
@Joozty this does not communicate to the user at all. I hope it will keep _Loading..._ forever because of `undefined` (nullish) values inside `crossDepositLag` property. Although, infinite loading seems to be fine to me at the moment.

I do not need more than this at this point for #335 but feel free to properly solve the bug.

At the moment I see 2 options:

1. Add status processing to `Root.defaultOptions.request` and if it's 202, return another promise that waits some time and repeats the request. It will keep the infinite loading and will not communicate the state properly but gives a hope that the request finishes sometime.

2. Throw 202 as some kind 'AwaitingError' and check it in place and communicate to the user. Also, applying the modified first option is possible.

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -65,13 +65,13 @@ const processStatus = (response) => {
 }
 
 const processBody = (response, { method }) => {
-  const { headers } = response
+  const { status, headers } = response
   const type = headers.get('Content-Type')
 
   return (/application\/([\w.-]\+)?json/g.test(type) && method !== 'HEAD'
     ? response.json()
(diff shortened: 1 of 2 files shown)


viktor-yakubiv:
I would like to merge this PR for now and solve the problem later.

About the actual solution, I prefer a simpler solution first, actually the first. If it's not enough, we may adapt the second solution, when we get there.

What do you think, @Joozty?

PR #341, opened 2020-05-21
Fix misspelled Url suffix

viktor-yakubiv:
This is a bulk rename `UrL` to `Url`. Waiting for Vercel to deploy and check it works 🙂


viktor-yakubiv:
I do not see any problems with this so it can be merged

Issue #342, opened 2020-05-22
Fix history loop


Joozty:
When user try to navigate back on overview tab it redirects them back to the overview table because we use router push instead of `replace` in [here](https://github.com/oacore/dashboard/blob/master/pages/_app/hooks.js#L15)

See video for more info 

https://www.loom.com/share/d7c819cebddf438db50b8447c7f5dd18

PR #343, opened 2020-05-22
Fix history loop

Joozty:
Closes https://github.com/oacore/dashboard/issues/342

### DIFF
--- a/pages/_app/hooks.js
+++ b/pages/_app/hooks.js
@@ -3,7 +3,11 @@ import { join } from 'path'
 import { useEffect } from 'react'
 import { useRouter } from 'next/router'
 
-const useRedirect = (routePath, actualPath = routePath) => {
+const useRedirect = (
+  routePath,
+  actualPath = routePath,
+  { replace = true } = {}
+) => {
   const router = useRouter()
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


pages/_app/hooks.js
```diff
@@ -3,7 +3,7 @@ import { join } from 'path'
 import { useEffect } from 'react'
 import { useRouter } from 'next/router'
 
-const useRedirect = (routePath, actualPath = routePath) => {
+const useRedirect = (routePath, actualPath = routePath, replace = true) => {
```
viktor-yakubiv, line 6: Please, avoid boolean parameters without a name, it's not obvious. Imagine the call like `useRedirect('/hello', '/hello', false)`. What does it mean?

Prefer options instead: `useRedirect('/hello', '/hello', { replace: false })`, it looks better. However, repeated parameter value does not seem to be good still.

In this case, just do not add a parameter. We do not need it now and I am sure we will never need it.

Joozty: Doesn't your editor show you the parameter name?

viktor-yakubiv: My one does not. I do no like that JetBrains' IDEs doing this because I used to know how long is my code (in columns) and it's annoying I cannot do it.

Also, GitHub does not do it, so for code review, it would be better to have a named boolean parameter. This is a good practice.

viktor-yakubiv: Despite, I may overuse options. I wonder what you do think about.

Joozty: it makes sense to use them and you are right with Github. I am just lazy guy I would add it there once there is too many parameters.

PR #344, opened 2020-05-22
Show correct message when invitation cannot be send

Joozty:
@viktor-yakubiv The reason why I couldn't get an email is that there is a continuing CORS issue. 😢 Obviously this PR doesn't fix it. It just makes UI show correct message.

### DIFF
--- a/store/organisation.js
+++ b/store/organisation.js
@@ -16,7 +16,7 @@ class Organisation extends Store {
       if (error instanceof NotAcceptableError)
         return { message: 'Email for this organisation is already registered.' }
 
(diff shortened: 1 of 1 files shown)


store/organisation.js
```diff
@@ -15,6 +15,8 @@ class Organisation extends Store {
     } catch (error) {
       if (error instanceof NotAcceptableError)
         return { message: 'Email for this organisation is already registered.' }
+
+      return { message: 'Something went wrong. Please try it later.' }
```
viktor-yakubiv, line 19:
```suggestion
      return { message: 'Something went wrong. Please try again later.' }
```

PR #345, opened 2020-05-22
RIOXX card

Joozty:


components/stacked-vertical-bar-chart/styles.module.css
```diff
@@ -31,6 +31,12 @@
   white-space: nowrap;
 }
 
+.caption-value {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
```
Joozty, line 37: I have literally no idea how to make these ellipsis work. Don't you know any css trick?

s

viktor-yakubiv: I think you can avoid ellipsis here. You do not need it. Also, the right label may be redundant. Just add minimal width for the inner element to fit the text. It would be clear enough, I think.

I've been thinking about this last night, we should probably use `<progress>` and `<meter>` where it's possible.

Joozty: Didn't know about those elements. We should definitely use them. :)

Joozty: is it ok if caption won't be there at all? I guess there is no way to add caption inside of progress/meter

viktor-yakubiv: Sorry, it [was not a good idea](https://stackoverflow.com/a/41430146/7205866). We should probably stick with some kind of chart.

I will take a look at this during next week if I have time.

viktor-yakubiv: I will rewrite this to pure CSS and we will see how it looks

viktor-yakubiv:
I consider this part done. I wonder if you @mcancellieri and @PMRuma have any opinion about. Please, check the [preview](https://dashboard-git-riox-card.oacore.now.sh).

I am going to prettify and merge this tomorrow.

PR #348, opened 2020-05-27
Remove ORCID card placeholder

viktor-yakubiv:
Let's stop promising it!

### DIFF
--- a/templates/overview/index.jsx
+++ b/templates/overview/index.jsx
@@ -2,12 +2,7 @@ import React from 'react'
 import { classNames } from '@oacore/design/lib/utils'
 
 import styles from './styles.module.css'
-import {
-  PlaceholderCard,
-  DataStatisticsCard,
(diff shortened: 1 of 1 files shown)


templates/overview/index.jsx
```diff
@@ -51,7 +46,6 @@ const OverviewTemplate = ({
       enrichmentSize={doiEnrichmentSize}
       dataProviderId={dataProviderId}
     />
-    <PlaceholderCard title="ORCiDs" value={5.8} />
```
Joozty, line 54: You can remove PlaceholderCard component too. If we need it we can revert it later.

viktor-yakubiv: I would like to keep it at the moment. I am almost sure we will get a request to put it back or create another card.

PlaceholderCard will disappear when we move everything to Nivo charts because linter will point that recharts got removed.

PR #353, opened 2020-06-01
Add table config updates

viktor-yakubiv:
Add simultaneous updates of action, sidebar and columns with props.

Fixes the bug with empty sidebar if there is any `pages.error`. @Joozty I wonder why it happens, by the way?

### DIFF
--- a/components/table/table.jsx
+++ b/components/table/table.jsx
@@ -51,9 +51,7 @@ class InfiniteTable extends React.PureComponent {
       selectedOption: null,
     }
 
-    this.action = action
-    this.sidebar = sidebar
-    this.columns = columns
+    Object.assign(this, { action, sidebar, columns })
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


components/table/table.jsx
```diff
@@ -64,6 +62,11 @@ class InfiniteTable extends React.PureComponent {
     )
   }
 
+  componentDidUpdate() {
+    const { action, sidebar, columns } = this.getConfig()
+    Object.assign(this, { action, sidebar, columns })
+  }
```
viktor-yakubiv, line 68: @Joozty I wonder if you have time to review this. From what I read in the documentation, this method is not called before the `render()` and it's not guaranteed that `render()` gets updated fields.

However, this works for some reason. I do not know how to do this better.

Joozty: Let's rather merge this https://github.com/oacore/dashboard/pull/326. It's already finished and included a fix for this.

viktor-yakubiv: My roadmap is the following:
1. release the current state
2. merge IRUS (#317) and RIOXX (#345)
3. test and release once again

This bug is the only thing that prevents step 1 and the rest.

I am afraid to merge #326 because it could postpone releases. If this fix is okay, let's do this first since it's going to be removed anyway unless you think it's going to cause more bugs.

Joozty: I don't know much about the bug. As long as it fixes it. Feel free to merge.

PR #360, opened 2020-06-03
Improve DOI chart

viktor-yakubiv:
1. Adds overall progress bar instead of 2 bars. Simplifies the chart, uses pure CSS to display it.
2. Adds a number with enrichment and another with overall coverage.
3. Moves the export button to the bottom. Removes ExportCard.

@PMRuma could you please check how it looks?


viktor-yakubiv:
@Joozty I need your help here. Is it possible to reduce the initial table size to 10 items only?

Joozty:
@viktor-yakubiv have  a look [here](https://github.com/oacore/dashboard/blob/master/templates/deposit-compliance/cards/table-card.jsx#L89).

viktor-yakubiv:
@Joozty table footer disappeared after merging #326. Everywhere except the Issues section.

PR #367, opened 2020-06-04
Change the email and contact-reason for premium requests

viktor-yakubiv:


viktor-yakubiv:
@mcancellieri I used `SALES` as a product. Let me know if this should be changed.

PR #374, opened 2020-06-11
Fix applying Table.Body callbacks without a reason

viktor-yakubiv:
When a user clicks on the row in the table without Sidebar, empty space appears with no data.

This prevents such behaviour by passing callbacks conditionally.

### DIFF
--- a/components/table/table.jsx
+++ b/components/table/table.jsx
@@ -61,6 +61,12 @@ const InfiniteTable = ({
       )
   }, [])
 
+  const bodyHasCallbacks = Boolean(sidebar)
+
+  // even if this is not an actual Hook, it starts from 'use-'
+  // to potentially have useCallback() call behind
+  const useRowCallback = (handler) => (bodyHasCallbacks ? handler : null)
+
   return (
     <div
       ref={containerRef}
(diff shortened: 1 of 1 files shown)


components/table/table.jsx
```diff
@@ -61,6 +61,12 @@ const InfiniteTable = ({
       )
   }, [])
 
+  const bodyHasCallbacks = Boolean(sidebar)
+
+  // even if this is not an actual Hook, it starts from 'use-'
+  // to potentially have useCallback() call behind
+  const useRowCallback = (handler) => (bodyHasCallbacks ? handler : null)
```
Joozty, line 68: Why don't you add useCallback here?

viktor-yakubiv:
I am not very satisfied with the solution and wonder if you @Joozty have any idea about. Ideally, this check should be included in the hook but passing an extra parameter to the hook is not a good idea. Maybe the current solution is the most correct one for the current app state.

components/table/table.jsx
viktor-yakubiv: I was afraid to use it here maybe because of the boundary (which I did not check). Also, this is not needed at the moment. Currently, the function returns handler only when it's needed, it's a simple check.

PR #376, opened 2020-06-16
Use aliases for React and React-DOM

Joozty:
This ensures that @oacore/design doesn't use own react copy when using `npm link @oacore/design` command


viktor-yakubiv:
Is there no way to make this work so in the `@oacore/design`?

Joozty:
I couldn't find a solution it came to my mind too.

Joozty:
Merging this. If we find a better solution directly in design then it can be removed.

PR #378, opened 2020-06-18
Fix password reset

Joozty:


Joozty:
@viktor-yakubiv Any reason for using [label class instead of element selector](https://github.com/oacore/design/blob/master/src/elements/forms/text-field.css#L125)? I am not sure if the third commit shouldn't be rather fixed in design. Hmm?

viktor-yakubiv:
I should have work according to [this line](https://github.com/oacore/design/blob/master/src/elements/forms/text-field.css#L247) but I have no idea, what is the problem there.

Joozty:
I am ok with the solution. I was just curious. 🙂

viktor-yakubiv:
There something is wrong when the form is filled up. When you type password and press 'Tab', the label returns back. I am not sure it's either missing classes in the HTML or the bug in Design.

Joozty:
The third commit was supposed to fix it. I just forgot to fix login page.

viktor-yakubiv:
It's not a problem, it can be done separately

Joozty:
Can you please double-check on master branch?

PR #379, opened 2020-06-19
Polyfill ResizeObserver

Joozty:
Apparently [ResizeObserver is not present in Edge ](https://caniuse.com/#feat=resizeobserver)<=18.x.x hence this polyfill.

### DIFF
--- a/components/forms/iframe-form.jsx
+++ b/components/forms/iframe-form.jsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useCallback, useRef } from 'react'
+import ResizeObserver from 'resize-observer-polyfill'
 
 import styles from './styles.module.css'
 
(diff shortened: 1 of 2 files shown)


components/forms/iframe-form.jsx
```diff
@@ -1,4 +1,5 @@
 import React, { useEffect, useCallback, useRef } from 'react'
+import ResizeObserver from 'resize-observer-polyfill'
```
viktor-yakubiv, line 2: Can this be implicitly imported through Webpack somehow?

Joozty: I followed the recommended way https://www.npmjs.com/package/resize-observer-polyfill#usage-example

Joozty:
It did not change because some package already uses it.

PR #380, opened 2020-06-23
Export content

Joozty:


Joozty:
I removed it because there the value was not used anywhere. The export is [automatically disabled when there are no data](https://github.com/oacore/dashboard/blob/master/templates/deposit-compliance/cards/table-card.jsx#L98).

PR #381, opened 2020-06-23
Don't show logout message when /user endpoint is just checked for status

Joozty:
IMHO this message may confuse the user. There is a use case that for example user never logged in and they see this message which is wrong.


Joozty:
Yep, this message is only shown when the user manually clicks logout button.

PR #382, opened 2020-06-23
Use /user/data-providers endpoint

Joozty:
I hope I haven't missed anything. It's affecting the whole dashboard. So we should probably wait a little but and create a separate release. The biggest change is that data providers are now a full part of a user entity and changes on [the root store are reflected through the callback](https://github.com/oacore/dashboard/blob/data-providers/store/root.js#L109).

### DIFF
--- a/store/root.js
+++ b/store/root.js
@@ -12,8 +12,8 @@ import { AccessError, AuthorizationError, PaymentRequiredError } from './errors'
 import apiRequest from 'api'
 import * as NetworkErrors from 'api/errors'
 
-const REPEADTED_REQUEST_TIMEOUT = 30000
-const REPEADTED_REQUEST_LIMIT = 5
+const REPEATED_REQUEST_TIMEOUT = 30000
+const REPEATED_REQUEST_LIMIT = 5
 
 class Root extends Store {
   static defaultOptions = {
@@ -31,11 +31,11 @@ class Root extends Store {
             if (attemptCount < 2) this.requestsInProgress -= 1
           })
           .then((response) =>
-            response.status === 202 && attemptCount < REPEADTED_REQUEST_LIMIT
+            response.status === 202 && attemptCount < REPEATED_REQUEST_LIMIT
               ? new Promise((resolve, reject) => {
                   const repeatedRequest = () =>
                     requestContinuously(url, options).then(resolve, reject)
-                  setTimeout(repeatedRequest, REPEADTED_REQUEST_TIMEOUT)
+                  setTimeout(repeatedRequest, REPEATED_REQUEST_TIMEOUT)
                 })
               : response
           )
@@ -71,11 +71,9 @@ class Root extends Store {
 
   contactUrl = `${process.env.API_URL}/contact`
 
-  @observable user = null
+  @observable user = new User({}, this.options)
 
   @observable organisation = null
-
-  @observable dataProvider = null
 
   @observable irus = null
 
@@ -105,26 +103,27 @@ class Root extends Store {
   }
 
   @computed
-  get dataProviders() {
-    return this.user?.dataProviders
+  get dataProvider() {
+    return this.user.dataProvider
+  }
+
+  set dataProvider(id) {
+    this.user.dataProvider = id
   }
 
   @computed
-  get baseUrl() {
-    if (!this.dataProvider) return null
-    return `/data-providers/${this.dataProvider.id}`
+  get dataProviders() {
+    return this.user.dataProviders
   }
 
   @action async init(dataProviderId) {
     try {
-      this.user = new User({}, this.options)
       await this.user.retrieve()
     } catch (unauthorizedError) {
       if (!this.options.allowAnonymousAccess)
         throw new AuthorizationError('Anonymous users are not allowed')
     }
     this.organisation = new Organisation(this.user.affiliationUrl, this.options)
-
     this.organisation.retrieve()
 
     this.changeDataProvider(dataProviderId)
@@ -146,7 +145,6 @@ class Root extends Store {
     if (this.dataProvider != null && hasTargetId(this.dataProvider)) return
 
     // Check access rights
-    // TODO: Should be moved or ideally removed
     const dataProvider = this.dataProviders.find(hasTargetId)
     if (dataProvider == null) {
       const dpStr = `DataProvider#${id}`
@@ -154,6 +152,7 @@ class Root extends Store {
     }
 
     this.dataProvider = dataProvider
+
     this.reset()
     this.retrieveStatistics()
     this.retrievePluginConfig()
@@ -182,29 +181,6 @@ class Root extends Store {
     data.forEach((plugin) => {
       this.plugins[plugin.type] = plugin
     })
-  }
-
-  @action
-  async updateDataProvider(patch) {
-    try {
-      // TODO: Should be this.dataProvider.url
-      const url = `/data-providers/${this.dataProvider.id}`
-      const { data } = await apiRequest(url, {
-        method: 'PATCH',
-        body: patch,
-      })
-      Object.assign(this.dataProvider, data)
-    } catch (networkOrAccessError) {
-      // Ignore errors for this moment
-    }
-  }
-
-  @action
-  async updateOrganization(patch) {
-    const { name: institution } = patch
-    this.updateDataProvider({ institution })
-
-    // TODO: Should be a method without cross-call to another method
   }
 
   @action
@@ -240,6 +216,28 @@ class Root extends Store {
     }
   }
 
+  @action
+  async updateDataProvider(patch) {
+    try {
+      const url = `/data-providers/${this.dataProvider.id}`
+      const { data } = await this.options.request(url, {
+        method: 'PATCH',
+        body: patch,
+      })
+      Object.assign(this.user.dataProvider, data)
+    } catch (networkOrAccessError) {
+      // Ignore errors for this moment
+    }
+  }
+
+  @action
+  async updateOrganization(patch) {
+    const { name: institution } = patch
+    await this.updateDataProvider({ institution })
+
+    // TODO: Should be a method without cross-call to another method
+  }
+
   async sendContactRequest(data) {
     try {
       await this.request(this.contactUrl, { method: 'POST', body: data })
@@ -251,7 +249,7 @@ class Root extends Store {
   }
 
   @action
-  async reset() {
+  reset() {
     this.irus = null
     this.rioxx = null
   }
--- a/store/root.js
+++ b/store/root.js
@@ -107,8 +107,14 @@ class Root extends Store {
     return this.user.dataProvider
   }
 
-  set dataProvider(id) {
-    this.user.dataProvider = id
+  set dataProvider(dataProvider) {
+    // Check access rights
+    if (!this.user.canManage(dataProvider.id)) {
+      const dpStr = `DataProvider#${dataProvider.id}`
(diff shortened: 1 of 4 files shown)


Joozty:
The reason for moving data provider to user entity was just to keep consistency with API. data providers has been always part of the user and computed property was just a duplicate value in the store. Nothing really computed was there. If you insist I will revert it.

store/root.js
```diff
@@ -104,80 +104,45 @@ class Root extends Store {
...
+  @action handleDataProviderChange(dataProviderId) {
     this.reset()
-    this.retrieveStatistics()
-    this.retrievePluginConfig()
-    this.retrieveIrusStats()
-    this.retrieveRioxxStats()
+    this.retrieveStatistics(dataProviderId)
+    this.retrievePluginConfig(dataProviderId)
+    this.retrieveIrusStats(dataProviderId)
+    this.retrieveRioxxStats(dataProviderId)
```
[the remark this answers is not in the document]
Joozty: >  If you want to make a pure function, you should get the result here and store it yourself.

Not sure what did you mean by this.

viktor-yakubiv:
I do. In my roadmap (sorry, I did not communicate it to you) we fetch data provider entity straight from the API without checking the user entity. We can do it checking current page location and should do this for 2 reasons: optimisation and allowing admins to view any resource.

store/root.js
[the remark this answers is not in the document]
viktor-yakubiv: If the function is pure, it accepts parameters and returns the result without storing anything anywhere. If the function accepts parameters, get the data and store it in `this`, there is no reason to pass parameters from `this`. That is why I prefer computed property here (which is computed at the moment but can be fetched directly as [I said below](https://github.com/oacore/dashboard/pull/382#issuecomment-652854710)).

Joozty:
@viktor-yakubiv can you check it now, please?

store/root.js
```diff
@@ -105,56 +103,31 @@ class Root extends Store {
...
-    if (!this.dataProvider) return null
-    return `/data-providers/${this.dataProvider.id}`
+  get dataProviders() {
+    return this.user?.dataProviders
   }
 
   @action async init(dataProviderId) {
     try {
-      this.user = new User()
+      this.user.onDataProviderChange(this.handleDataProviderChange.bind(this))
```
viktor-yakubiv, line 117: I do not like fetching the user before calling `init()` method. That is why I did creating here.

If we have an init method for the initialisation, we should not run heavy operations like calling the API in the constructor.

store/user/dashboard.js
```diff
@@ -15,6 +25,58 @@ class DashboardUser extends User {
       (e) => e.name.toLowerCase().search(searchTerm.toLowerCase()) !== -1
     )
   }
+
+  onDataProviderChange(callback) {
```
viktor-yakubiv, line 29: The method is not clear, it's better to have `addEventListener` or `on`, or extend EventTarget somewhere and use its API. However, we can keep this for now.

store/user/dashboard.js
```diff
@@ -15,6 +25,58 @@ class DashboardUser extends User {
...
+
+    // Check access rights
+    const dataProvider = this.dataProviders.find(hasTargetId)
+    if (dataProvider == null) {
+      const dpStr = `DataProvider#${id}`
+      throw new AccessError(`${this.user} does not have access to the ${dpStr}`)
+    }
+
+    this.dataProvider = dataProvider
+    this.#onDataProviderChangeCallback.forEach((callback) => callback())
```
viktor-yakubiv, line 56: Prefer `dispatchEvent` here.

store/user/dashboard.js
```diff
@@ -15,6 +25,58 @@ class DashboardUser extends User {
...
+    }
+  }
+
+  @action
+  async updateOrganization(patch) {
+    const { name: institution } = patch
+    this.updateDataProvider({ institution })
+
+    // TODO: Should be a method without cross-call to another method
+  }
```
viktor-yakubiv, line 79: This still should not be a part of the `User` entity.

store/root.js
Joozty: I am confused. API is not called in constructor. User and dataproviders are loaded when `.retrieve()` method is called. We don't use `new User({prefetch: true})`.

viktor-yakubiv: Sorry, it's my fault.

store/user/dashboard.js
Joozty: If you wish we can create some global store event bus.

store/user/dashboard.js
Joozty: https://github.com/oacore/dashboard/pull/382#discussion_r448888015

store/user/dashboard.js
viktor-yakubiv: I wanted to but I did not because it was not necessary

Joozty: All right I think it will be better with the event system. Let me change it.

store/user/dashboard.js
Joozty: So you would leave it in root?

Joozty: I think it stay as is for the time being. We need a new DataProvider/Organisation substore.

store/user/dashboard.js
Joozty: [`a3d3422` (#382)](https://github.com/oacore/dashboard/pull/382/commits/a3d342287a793bd52fe3f55652efe3339bce68e8)

store/user/dashboard.js
viktor-yakubiv: I do not understand the reason for moving it here. It was designed as desired before. The data provider is a member of the application model but not of the user. The application controls all changes/switches, updates, loads and unloads stuff. If the user can discover it's own data providers it's exposed to the application and the application decides based on the user's list should the user access the data provider or not.

Joozty:
I don't remember the reason now since this or is quite old. I think I had a reason. Maybe just it was because I wanted to structure our store the same way as the API is structured, i.e. data providers are exposed under /user path. 🤷‍♂️

store/user/dashboard.js
Joozty: Isn't it kind of breaking single responsibility thing? We were trying to change user property outside of user entity? I don't know and I truly I don't see a big difference in either.

viktor-yakubiv: We don't. The _User_ has a list of things it can access. The _Application_ is guided (read-only) by this list but still decides what to do.

Furthermore, the _User_ is a part of the _Application_, it should be simple and the _Application_ should be more complex.

I do not think this has anything to the SRP. The SRP claims that the class should be changed because of one and only one actor. I am struggling to define actors here but my guess is that's API for the _BaseUser_ and the Dashboard application for the _DashboardUser_.

Joozty: Ohh I called it wrong I guess. My point was to change things only inside of the object. Anyway so is this something why the PR cannot be merged? Because you are giving me only comment reviews so I have no idea what I am supposed to do.

Joozty: Hope it's now ok with you.

store/root.js
```diff
@@ -105,26 +103,33 @@ class Root extends Store {
...
   @computed
-  get dataProviders() {
-    return this.user?.dataProviders
+  get dataProvider() {
+    return this.user.dataProvider
+  }
+
+  set dataProvider(dataProvider) {
+    // Check access rights
+    if (!this.user.canManage(dataProvider.id)) {
```
viktor-yakubiv, line 112: Here should be another condition kind of `!this.options.allowAnonymousAccess && ...` but it can be added in the future.

Also, it would be good it the `use.canManage` check the user's permissions, it the user is an admin, it should always return `true`

store/root.js
```diff
@@ -137,23 +142,15 @@ class Root extends Store {
       return
     }
 
-    // Compare strings and numbers
-    // since we do not really know what type of ID the API exposes
-    // eslint-disable-next-line eqeqeq
-    const hasTargetId = (dataProvider) => id == dataProvider.id
+    const dataProviderInt = parseInt(id, 10)
```
viktor-yakubiv, line 145: Why don't you like it? I added a function like this because the API really can migrate IDs to UUIDs (@samuelpearce tends to say that, let me know if I am wrong), what will make use to change this once again.

store/root.js
```diff
@@ -137,23 +142,15 @@ class Root extends Store {
...
+    if (this.dataProvider?.id === dataProviderInt) return
 
-    // Check access rights
-    // TODO: Should be moved or ideally removed
-    const dataProvider = this.dataProviders.find(hasTargetId)
-    if (dataProvider == null) {
-      const dpStr = `DataProvider#${id}`
-      throw new AccessError(`${this.user} does not have access to the ${dpStr}`)
-    }
+    this.dataProvider = this.user.getDataProviderById(dataProviderInt)
```
viktor-yakubiv, line 152: We probably should start fetching data providers directly from the API, not from the user. Let's think about on the next step.

store/user/dashboard.js
```diff
@@ -4,12 +4,21 @@ class DashboardUser extends User {
...
+  retrieve(...scopes) {
+    return super.retrieve('dataProviders', ...scopes)
   }
 
   canManage(dataProviderId) {
     return this.dataProviders.some(({ id }) => dataProviderId === id)
   }
 
+  getDataProviderById(dataProviderId) {
+    return this.dataProviders.find(({ id }) => dataProviderId === id)
```
viktor-yakubiv, line 19: A similar remark here. I would use `==` but it's not a big deal.

store/user/dashboard.js
```diff
@@ -4,12 +4,21 @@ class DashboardUser extends User {
   static defaultValues = {
     ...User.defaultValues,
     dataProviders: [],
+    dataProvider: null,
```
viktor-yakubiv, line 7: ~You don't change the data provider here, do you?~ You do, I see.

It's kind of okay because you check access rights in the setter, what is good. However, storing the parent's data in the child object does not feel good. You can add a private variable to the `Root` or rethink data provider changing action.

store/root.js
```diff
@@ -240,6 +214,28 @@ class Root extends Store {
     }
   }
 
+  @action
```
viktor-yakubiv, line 217: You changed the order of the methods. It does not seem you had a reason for this, can you move this block above the `retrieveIrusStats()`?

store/user/dashboard.js
Joozty: > However, storing the parent's data in the child object does not feel good.

That's why I wanted to have everything in user entity.

store/root.js
Joozty: Actually I had a reason. changeDataProvider method calls retrieveX methods so I wanted group them together and place them directly bellow that method. :)

store/root.js
Joozty: I am not saying that I don't like it. I don't think about these things honestly. I found it better to convert it because I spared one or two lines of code. That's all. If dataProvider changes to UUID I am happy to change it. :)

store/user/dashboard.js
viktor-yakubiv: You can if you want to. JavaScript allows it. However, I would move it to the parent later.

store/root.js
Joozty: Yeah truth but allowAnonymousAccess doesn't work yet. So it can be added later as you said. The same for the second paragraph.

PR #383, opened 2020-06-23
Show more detailed error messages on registration

Joozty:


public/secure/invitation.js
```diff
@@ -22,8 +24,12 @@ function performApiRequest(url, data) {
       throw new UnauthorisedError('Wrong credentials')
     }
 
-    // TODO: We should distinguish errors.
-    //     Currently API sends only 400 `{"message": "error"}`
+    if (response.status === 406)
+      throw new InvitationCodeInvalid('Invitation code is invalid.')
+
+    if (response.status === 409)
+      throw new InvitationCodeAlreadyUsed('This code was already used.')
```
viktor-yakubiv, line 31: If this message is shown to the user and it's actually about the registration of the invited user, please add the following change.

```suggestion
      throw new InvitationCodeAlreadyUsed('Seem like you have registered already. Please try to log in.')
```

PR #384, opened 2020-07-03
Migrate Organisation to Resource

Joozty:


store/resource.js
```diff
@@ -4,7 +4,7 @@ import apiRequest from 'api'
 
 class Resource {
   static defaultOptions = {
-    request: (...args) => apiRequest(args).then(({ data }) => data),
+    request: (...args) => apiRequest(args).then((data) => data),
```
viktor-yakubiv, line 7: It looks like you don't need a repeating function here.

```suggestion
    request: (...args) => apiRequest(args),
```

And now it looks like you do not need to override `request` at all, do you?

```suggestion
    request: apiRequest,
```

Unless you want to bind the `apiRequest`. However, it does not accept the context.

store/resource.js
```diff
@@ -42,7 +42,7 @@ class Resource {
   retrieve(...scopes) {
     const { request } = this.options
     if (this.id == null) {
-      return request(this.url).then((data) => {
+      return request(this.url).then(({ data }) => {
```
viktor-yakubiv, line 45: Why do you do this? Do you have any place to check headers or status?

store/root.js
```diff
@@ -117,16 +117,18 @@ class Root extends Store {
...
     try {
-      this.user = new User()
+      this.user = new User({}, this.options)
       await this.user.retrieve()
     } catch (unauthorizedError) {
       if (!this.options.allowAnonymousAccess)
         throw new AuthorizationError('Anonymous users are not allowed')
     }
+    this.organisation = new Organisation(
+      `/organisations/${this.user.organisationId}`,
```
viktor-yakubiv, line 127: The User has an `affiliationUrl` what is the actual user's organisation. Use that property instead.

```suggestion
      this.user.affiliationUrl,
```
If you accept, this will lead to Prettier error.

store/resource.js
Joozty: Our [store request](https://github.com/oacore/dashboard/blob/master/store/root.js#L22) works like that. So I just wanted to unify it.

store/root.js
Joozty: Actually I guess it cannot be used if we allow anonymous access. this.user will be undefined.

Joozty: oh never mind. Stupid comment. 😄

PR #386, opened 2020-07-14
Fix dataprovider

Joozty:
Apparently some people try to access data providers not assigned to them. First commit fixes the issue 
`Cannot read property 'id' of undefined` but it's not a really fix for the issue itself. 

In the second commit I tried to cover case when user access data provider which is not managed by them. If it happens user is redirected to `/data-providers` which redirect them to the fist available data provider. It's probably better than blank screen.  I am not entirely sure about the second commit. I had to do some changes I sort of don't  like but there was no other option I could think of.

Based on https://sentry.io/organizations/oacore/issues/1781836363/?project=2091955&query=is%3Aunresolved

### DIFF
--- a/pages/_app/index.jsx
+++ b/pages/_app/index.jsx
@@ -6,7 +6,7 @@ import '@oacore/design/lib/index.css'
 import './global.css'
 
 import { UnauthorizedError } from 'api/errors'
-import { AuthorizationError } from 'store/errors'
+import { AuthorizationError, AccessError } from 'store/errors'
 import { logPageView } from 'utils/analytics'
 import { initStore, GlobalProvider } from 'store'
 import Application from 'components/application'
@@ -100,7 +100,7 @@ class App extends NextApp {
       pathname.match(/\/data-providers\/([^/]+)\/?/) ?? []
     const { store } = this
     store
-      .init(dataProviderId)
+      .init()
       .then(() => {
         Sentry.configureScope((scope) => {
           scope.setUser({
@@ -110,9 +110,11 @@ class App extends NextApp {
         })
 
         // Subscribe store to router after initialization
-        router.events.on('routeChangeStart', this.handleRouteChange)
+        // We use routeChangeComplete otherwise router.query contains previous
+        // values
+        router.events.on('routeChangeComplete', this.handleRouteChange)
 
-        // Trigger App render
+        store.changeDataProvider(dataProviderId)
         this.setState({ isAuthorized: true })
       })
       .catch((error) => {
@@ -121,6 +123,13 @@ class App extends NextApp {
           this.redirectToLogin({
             reason: '',
           })
+        } else if (error instanceof AccessError) {
+          // If the state is updated before route changes NextJS may render
+          // the old component. Since we don't have any data for that provider
+          // it may fail. (Works, DepositDates are not initialized...)
+          router
+            .replace('/data-providers')
+            .then(() => this.setState({ isAuthorized: true }))
         } else throw error
       })
   }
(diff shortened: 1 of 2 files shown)


viktor-yakubiv:
> case when user access data provider which is not managed by them. If it happens user is redirected to /data-providers which redirect them to the fist available data provider. 

It should display an error page. I wonder if Next allows redirecting to the `_error` with 403 status code and maybe another parameter for the scope but it may not be needed.

I think for such data providers the `dataProvider` to either `null`, an instance of the `Error` or an object like `{ error: theAccessError }`. It can be done in another PR or handled here, the contract is up to you, to me all of them are the same bad at the moment 😁

I will review a bit later

pages/_app/index.jsx
```diff
@@ -110,12 +110,21 @@ class App extends NextApp {
...
+        // values
+        router.events.on('routeChangeComplete', this.handleRouteChange)
+
+        try {
+          store.changeDataProvider(dataProviderId)
+        } catch (error) {
+          if (error instanceof AccessError)
+            await router.replace('/data-providers')
+          else throw error
+        }
```
viktor-yakubiv, line 123: If I understand this correctly, when `then`'s callback throws, the error is caught but the following `catch`. You can handle all the errors in a single place.

Joozty: Check it now.

Joozty:
> I like the code overall but I did not check how it works. Should I?

Let's hope I tested it correctly

From all options 404 page would suit us best. With the upcoming new version of @oacore/design I will add it.

pages/_app/index.jsx
```diff
@@ -110,17 +110,22 @@ class App extends NextApp {
...
+        // We use routeChangeComplete otherwise router.query contains previous
+        // values
+        router.events.on('routeChangeComplete', this.handleRouteChange)
 
-        // Trigger App render
+        store.changeDataProvider(dataProviderId)
         this.setState({ isAuthorized: true })
       })
-      .catch((error) => {
+      .catch(async (error) => {
```
viktor-yakubiv, line 120: I still don't like it async, let see what we can do 🤔

pages/_app/index.jsx
```diff
@@ -110,17 +110,22 @@ class App extends NextApp {
...
         if (error instanceof AuthorizationError) {
           this.setState({ isAuthorized: false })
           this.redirectToLogin({
             reason: '',
           })
+        } else if (error instanceof AccessError) {
+          await router.replace('/data-providers')
+          this.setState({ isAuthorized: true })
```
viktor-yakubiv, line 128: What do you wait for the state changing? I will cause extra rendering and may cause extra redirecting. I am not sure this is what you want.

I would do it like this:
```js
          router.replace('/data-providers')
          this.setState({ isAuthorized: true })
```

or like this:
```js
          this.setState({ isAuthorized: true }, () => router.replace('/data-providers'))
```

What you did is like

```js
          router.replace('/data-providers').then(() => this.setState({ isAuthorized: true })
```

what is also fine but I am wondering why the order should be like this.

pages/_app/index.jsx
Joozty: You are very old school guy you know that? 😄 I surely noticed you prefer .then .catch instead of async/await and was waiting for this comment to come.

pages/_app/index.jsx
Joozty: I should have probably write a comment there it's tricky. Sorry about that. 

If the state is updated before route changes nextjs may render the old component. Since we don't have any data for that provider it will fail with error because Works, DepositDates are not initialized.

Joozty: It's very unlikely because the production build is really fast. Nonetheless it should be covered.

PR #387, opened 2020-07-14
Show error message when update organisation returns error

Joozty:
Based on https://sentry.io/organizations/oacore/issues/1783820044/?project=2091955&query=is%3Aunresolved

I notified @valerapro about the issue. In the meantime we should cover it on our side too.

### DIFF
--- a/components/forms/shell.jsx
+++ b/components/forms/shell.jsx
@@ -14,11 +14,14 @@ const FormShell = ({
   const [isChanged, setChanged] = useState(isButtonVisible)
   const [isSubmitted, setSubmitted] = useState(false)
 
-  const handleChange = () => setChanged(true)
+  const handleChange = () => {
+    setChanged(true)
+  }
 
   const handleSubmit = (event) => {
     setSubmitted(true)
     if (onSubmit) onSubmit(event)
+    event.target.reset()
   }
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


Joozty:
apart from the bug we discussed on Slack it's ready. I guess it's a not a big deal. :)

components/forms/shell.jsx
```diff
@@ -11,14 +11,18 @@ const FormShell = ({
...
 
-  const handleChange = () => setChanged(true)
+  const handleChange = () => {
+    setChanged(true)
+  }
 
   const handleSubmit = (event) => {
     setSubmitted(true)
     if (onSubmit) onSubmit(event)
+    formRef.current.reset()
```
viktor-yakubiv, line 25: Why don't you use `event.target` here? It will reduce a few lines of code.

store/root.js
```diff
@@ -223,15 +223,20 @@ class Root extends Store {
...
     } catch (networkOrAccessError) {
-      // Ignore errors for this moment
+      return {
+        message: 'Something went wrong. Please try it again later!',
+      }
     }
   }
 
   @action
   async updateOrganization(patch) {
```
viktor-yakubiv, line 237: If you return a promise, you don't need an `async` keyword here.

PR #388, opened 2020-07-16
Remove duplicate styles

Joozty:


components/application/head.jsx
```diff
@@ -21,15 +21,6 @@ const Head = () => (
       type="image/svg+xml"
       href="/favicon/favicon.svg"
```
Joozty, line 22: @viktor-yakubiv I am thinking about resolving all favicon assets from core.ac.uk. This way we can avoid having duplicate assets in all our projects.

PR #393, opened 2020-07-22
Bump @oacore/design

Joozty:


viktor-yakubiv:
I am thinking about creating HTML templates for pure HTML cases. Do you have any preference in the templating engine?

Joozty:
Not sure if we need any template engine. I don't have any experience in these things hence no preference.

viktor-yakubiv:
I have some experience and I did not like any, so I have no preference too 🙂

PR #395, opened 2020-07-27
Create dataProvider substore

Joozty:


store/root.js
```diff
@@ -149,67 +131,13 @@ class Root extends Store {
...
-      this.rioxx = {
-        partiallyCompliantCount,
-        compliantCount,
-        totalCount,
-        ...rest,
-      }
-    } catch (networkOrAccessError) {
-      // Ignore errors for this moment
-    }
+    this.dataProviderData = new DataProvider(
```
Joozty, line 134: I ran into name conflict hence this name. Maybe it would be worth to rename this.dataProvider -> this.activeDataProvider and this.dataProviderData -> this.dataProvider 🤔

store/root.js
```diff
@@ -149,67 +131,13 @@ class Root extends Store {
...
+      {
+        id: this.dataProvider.id,
```
Joozty, line 136: The question is whether DataProvider entity should contain all metadata like location etc. If so we may have duplicate data in User and DataProvider.

store/root.js
viktor-yakubiv, line 134: If you want to have a setter (what you do), you should have a getter too.

It's better to have it under `this.#dataProvider` with `get dataProvider()`. The current name looks confusing. The same applies to the templates.

---

However, I don't think it should be allowed to set the `dataProvider` to a string/number and get an object. If you want to do it right, create private property, a getter and a method like `changeDataProvider` that does all the checks and accepts ID. Otherwise, just use `dataProvider` which can be set to a DataProvider instance by a convention.

pages/data-providers/[data-provider-id]/content.jsx
```diff
@@ -3,11 +3,11 @@ import React from 'react'
 import { withGlobalStore } from 'store'
 import ContentTemplate from 'templates/content'
 
-const Content = ({ store, ...props }) => (
+const Content = ({ store: { dataProvider }, ...props }) => (
   <ContentTemplate
-    exportUrl={store.works.contentExportUrl}
-    works={store.works}
-    changeVisibility={store.works.changeVisibility}
+    eportUrl={dataProvider?.works.contentExportUrl}
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/content.jsx:8]:
```suggestion
    exportUrl={dataProvider?.works.contentExportUrl}
```

Joozty:
It's gonna be browser cache. We don't store it anywhere.

PR #437, opened 2020-08-05
Select from @oacore/design

Joozty:


package.json
```diff
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "@mdi/svg": "^5.4.55",
-    "@oacore/design": "^0.3.4",
+    "@oacore/design": "^1.0.0",
```
viktor-yakubiv, line 39: Wow! Unexpected

PR #439, opened 2020-08-07
Add Tooltips and adjust texts

Joozty:
Hopefully, I covered the vast majority of tooltips from document.

### DIFF
--- a/texts/depositing/data-overview.md
+++ b/texts/depositing/data-overview.md
@@ -1,4 +1,6 @@
 ---
 title: Data overview
 redirect: Review data
+tooltip-non-compliant: The number of documents that were deposited outside the 92 day window. Data is calculated from the deposit date that CORE collects from the metadata record and from the publication date, which is a combination from other repository data and Crossref.
+tooltip-outputs-counted: Number of outputs that are within this REF 2021 period, for which CORE has data and can calculate compliance. This number is used to create the compliance percentage of a repository. 
 ---
--- a/texts/plugins/overview.yml
+++ b/texts/plugins/overview.yml
@@ -6,5 +6,5 @@ discovery:
 recommender:
   title: Recommender
   description: |
-    Enrich your full text coverage dynamically adding missing bits from CORE
+    Enrich your full text coverage dynamically by adding missing content from CORE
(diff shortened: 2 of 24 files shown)


texts/plugins/overview.yml
```diff
@@ -6,5 +6,5 @@ discovery:
 recommender:
   title: Recommender
   description: |
-    Enrich your full text coverage dynamically adding missing bits from CORE
+    Enrich your full text coverage dynamically by adding missing content from CORE
```
Joozty, line 9: This description doesn't make sense to me. How do you enrich the full text coverage with Recommender? 🤔 And what missing bits does recommender add?

viktor-yakubiv: I edited the text without paying much attention to it, it was _CORE services: CORE Recommender and CORE Discovery_ before. I wanted to shorten it.

You can switch 2 names, write _CORE Discovery and CORE Recommender_, would it be more clear?

texts/depositing/data-overview.md
```diff
@@ -1,4 +1,6 @@
 ---
 title: Data overview
 redirect: Review data
+tooltip-non-compliant: The number of documents that were deposited outside the 92 day window. Data is calculated from the deposit date that CORE collects from the metadata record and from the publication date, which is a combination from other repository data and Crossref.
```
viktor-yakubiv, line 4: I don't like the speech from the third person but it is endless. I hope we will sort it out soon and will update the messages properly.

texts/plugins/overview.yml
Joozty: I am not following. 😕

viktor-yakubiv: I edited the string in the file where you go it but did not pay much attention to the result. I may be mistaken because I see only part of the sentence.

Joozty: What about something like `Get instant recommendations from CORE's database and discover articles of interest from across the network of open access repositories.`

viktor-yakubiv: `... and journals.` Sounds good!

PR #466, opened 2020-09-02
Fix image overflows card

Joozty:


Joozty:
The image look is improved here https://github.com/oacore/next/pull/247

PR #469, opened 2020-09-08
Allow to see any repository for super admin user

Joozty:


store/root.js
```diff
@@ -114,16 +114,32 @@ class Root extends Store {
...
 
     // Check access rights
     if (
       dataProviderMetaData !== null &&
       !this.user.canManage(dataProviderMetaData?.id)
     ) {
-      // TODO: allow access to any dataProvider for admin users
-      const dpStr = `DataProvider#${dataProviderMetaData?.id}`
-      throw new AccessError(`${this.user} does not have access to the ${dpStr}`)
+      if (!this.user.superAdmin) {
```
viktor-yakubiv, line 124: This should be on a different level. A bit above ☝️

`User.canManage(id)` should return `true` for any ID if the user is a super-admin. The rest of the code should stay generic.

viktor-yakubiv (changes requested):

store/root.js
Joozty: > This should be on a different level. A bit above ☝️

So in that case `getDataProviderById` should fetch dataProvider metadata if user is superAdmin?

> The rest of the code should stay generic.

What do you mean by that?

viktor-yakubiv: > So, in that case, `getDataProviderById` should fetch dataProvider metadata if user is `superAdmin` and data is missing?

I mean, that the if there is and `/data-providers/{id}`, the API call for the metadata goes anyway. The user who opens the URL most likely has an access to it. The `/user` call goes asynchronously too.

When both are retrieved, we can check the `canManage` condition.

`getDataProviderById` perhaps can be removed.

Joozty: There is no point in rerequesting the data we already have.

viktor-yakubiv: Let's say, there should be a method that creates a DataProvider store from the ID. Calling this method means that the user can manage the store.

The method itself can find a data provider in the user entity or retrieve it from the server, hence it should be async.

Joozty: If I understood you correctly basically the code I added should be just moved to a new method but the whole logic remains the same.

viktor-yakubiv: The `canManage` method should change to the following:

```js
  canManage(dataProviderId) {
    return this.superAdmin || this.dataProviders.some(({ id }) => dataProviderId === id)
  }
```

and you should utilise it within the Root store without relying on specific knowledge about the entities like having a `superAdmin` property in the User.

This particular `if` statement should not exist but the `changeDataProvider` method should be updated appropriately considering that some data can be missing.

Also, I don't think the current data provider should be added to the user's list but it should be shown somehow in the menu (as the first item, perhaps).

Joozty: Fair enough. I would suggest merging this PR as I don't see the things you are proposing as a blocker whatsoever. If you feel like it can be improved in any way feel free to refactor it. What do you think would that work for you? 🙂

viktor-yakubiv: Yeah, both options have pros and cons. I lost error reporting for 404 at `/data-providers/{id}`

pages/data-providers/[data-provider-id]/overview.jsx
```diff
@@ -5,22 +5,22 @@ import { withGlobalStore } from 'store'
 
 const Overview = ({ store, ...props }) => (
   <OverviewTemplate
-    metadataCount={store.dataProvider?.statistics.metadataCount}
-    fullTextCount={store.dataProvider?.statistics.fullTextCount}
-    timeLagData={store.dataProvider?.depositDates.timeLagData}
+    metadataCount={store.dataProvider?.statistics?.metadataCount}
```
Comment (Joozty) [inline pages/data-providers/[data-provider-id]/overview.jsx:8]: you will need to change everywhere not just in overview template

store/root.js
```diff
@@ -85,7 +85,14 @@ class Root extends Store {
...
   @computed
   get dataProviders() {
-    return this.user.dataProviders
+    const ids = new Set()
+    return [this.dataProvider, ...this.user.dataProviders].filter(
+      (dataProvider) => {
+        if (dataProvider == null || ids.has(dataProvider.id)) return false
+        ids.add(dataProvider.id)
+        return true
+      }
```
Joozty, line 94: Why do you filter it?

viktor-yakubiv: Due to implementation of the `changeDataProvider`, `this.dataProvider` can be `null`. I filter it to avoid an if-statements to check it with the following array unshifting.

pages/data-providers/[data-provider-id]/overview.jsx
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/overview.jsx:8]: This is what I am afraid of

store/root.js
```diff
@@ -100,54 +107,51 @@ class Root extends Store {
...
     })
   }
 
+  findDataProvider(id) {
+    const dataProvider = this.user.dataProviders.find(
+      // Comparing string ID from the URL with whatever it is in the object
+      // eslint-disable-next-line eqeqeq
+      ({ id: dataProviderId }) => dataProviderId == id
+    )
+    return dataProvider ?? { url: `/data-providers/${id}` }
```
Joozty, line 143: you need to fetch new repo metadata hand append them to all data providers array, if it is work in progress sorry abou my review.

viktor-yakubiv: Why do you think so?

store/root.js
```diff
@@ -100,54 +107,51 @@ class Root extends Store {
...
-        this.user.dataProviders = [
-          ...this.user.dataProviders,
-          dataProviderMetaData,
-        ]
-      } catch (error) {
-        throw new AccessError(
-          `Failed to load DataProvider#${id} metadata. Server responded with ${error.status}`
-        )
-      }
+    if (!this.user.canManage(id)) {
```
Joozty, line 125: If you allow all repositories for super admin (even the non-existing ones) you will never get error reporting. 

If you want to see 404 page you need to throw an error here hence I was fetching repository metadata in this method and thrown exception if it returned any status code but 200.

store/root.js
Joozty: Ok that makes sense but what about the other filtering. Is there a possibility that you may have two data providers with the same id?

store/root.js
Joozty: because the repository select works like that. 

or you will need to play with observables a little bit. Something like this (I haven't tested it so I may be talking  a bulshit 😄 )


```
class DataProvider extends Resource {
  @observable id = -1

  @observable name = 'Unknown repository'
```

viktor-yakubiv: Yes, this solves the problem with observing the value. Thank you

pages/data-providers/[data-provider-id]/overview.jsx
Comment (viktor-yakubiv) [reply pages/data-providers/[data-provider-id]/overview.jsx:8]: The only Settings page is not covered. This is what I don't cate much but we should keep it in mind. Also, the Settings page is not moved to templates.

store/root.js
viktor-yakubiv: This is what `ids.add(id)` stands for. If the set above already contains an ID, it's not added again. It's a trivial deduplication algorithm.

store/root.js
viktor-yakubiv: Non-existing data provider will get infinite loading or something. It will try to solve it but I think it's not bad as it is.

store/root.js
Joozty: Yep I understand the algorithm I was asking more why you added it there. Why are you worried about duplicate values?

viktor-yakubiv: I assume that it may happen that the user's data providers are not loaded yet and the direct request was triggered, so there may be a duplicated repository

store/root.js
Joozty: The current implementation redirects to the first existing data provider and I would say it should behave the same way.

viktor-yakubiv: You will not like the implementation (d520f7c), I guess :) Let me know, what you think

pages/_app/index.jsx
```diff
@@ -84,6 +84,10 @@ class App extends NextApp {
       // Don't report to console
       event.preventDefault()
     }
+
+    // Forcing reload because Next.js subscribes to this even first
+    if (event.reason instanceof NotFoundError)
```
Joozty, line 89: This sadly is very generic and may have an unintended effect.  You should probably create new exception like `DataProviderNotFound` or why don't you make changeDataProvider async so it can be caught correctly?

viktor-yakubiv: I am not sure, making something to wait for this method is a good idea.

The flow should be: start changing, check conditions, create an event (error/success).

Joozty: That's fine. So in that case, if you can please create a new exception and don't rely on NotFoundError because it can be thrown by other API call too.

viktor-yakubiv: Yes, I will

viktor-yakubiv: @Joozty do you think we need a specific error or generic one is enough as it is now: yet another NotFoundError but from the store, not from the API?

Currently, it may be too generic since it's triggered only in a single specific case when the data provider is not found.

viktor-yakubiv:
I created another rebased branch, that is called `superadmin-rebased`. @Joozty if you agree with the solution, I will push the branch here, merge and clean-up.

store/resource.js
```diff
@@ -10,6 +10,8 @@ class Resource {
 
   static defaultValues = {}
 
+  @observable error = null
```
Joozty, line 13: Why do you need thus property?

pages/_app/index.jsx
Joozty: I would say it would be better if it is specific something like `DataProviderNotFound`

Joozty:
I am ok with the solution. I added some comments here and https://github.com/oacore/dashboard/commit/e91be3973e6bdef77a5a94384370069e2e7a00c0 but those are minor. Feel free to solve them or merge without doing nothing. Thanks

store/resource.js
viktor-yakubiv: I don't but I thought this may be good if the method remains consistent, i.e. in the success scenario it stores some properties into `this`, so in the error scenario it stores the error into `this`.

Also, I think that the information itself may be useful too to show some error reporting in the future.

What do you think about?

Joozty: I try to follow my personal rule - everything which is not used shouldn't be in the repository.  It just makes the other developers think why is it there.  There has to be some reason and if it gets remove something will break.

viktor-yakubiv:
@mcancellieri do you want to test this?

PR #472, opened 2020-09-14
Add 404 page

Joozty:
I was trying to fix this issue and decided finally to add custom 404 page (inspiration taken from [static website](https://github.com/oacore/next/blob/master/components/error.jsx))

https://sentry.io/organizations/oacore/issues/1716970829/?project=2091955&query=is%3Aunresolved

### DIFF
--- a/components/application/activity-select/index.jsx
+++ b/components/application/activity-select/index.jsx
@@ -2,7 +2,6 @@ import path from 'path'
 
 import React from 'react'
 import Link from 'next/link'
-import { useRouter } from 'next/router'
 
 import activities from '../activities'
 import styles from './styles.module.css'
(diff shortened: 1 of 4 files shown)


viktor-yakubiv (approved):
We should have different error pages for different cases (in the priority descending):
- access is forbidden to the data provider – 403
  - propose to select a data provider the user has access to
- access allowed but the page URL does not exist – 404
  - either redirect to the nearest overview
  - or propose to go to the nearest overview or select another data provider
- unknown request – 404
  - propose to select a data provider
  - if the user is not logged in, perhaps redirect to the login page
    - if this is not a good practice, show something generic

Currently, in this PR, we copied a public error page, what is a good start but not as good for the project as it could be. I checked [a few things for inspiration](https://medium.muz.li/14-error-pages-to-inspire-your-next-design-a44d95925dcc) and [design recommendations](https://uxdesign.cc/how-to-create-great-error-pages-b2de5e5daca8) and will think about it more.

For now, having this page is better than not having it. Good job! Thank you 👍

PR #473, opened 2020-09-14
T&C modal

Joozty:


Joozty:
I was thinking maybe we should allow the user to switch to a different data provider? From the accessibility point of view I am afraid we cannot exclude app bar from modal overlay. We would need to add another select component into the modal if that's required. WDYT @viktor-yakubiv?

viktor-yakubiv:
Actually, because of this, I think that the terms and conditions should be rendered and shown on the separate page rather than in modal.

Another option is to allow closing the popup but showing an error on the overview tab that can be resolved only by accepting the terms.

The actual technical implementation relies much on the political decision which we should discuss. Let's ask @petrknoth  tomorrow.

Joozty:
@viktor-yakubiv Maybe we could merge this to the `toc` branch and make the final adjustments there. What do you think?

viktor-yakubiv:
I cannot open the preview, I get a blank screen. You can merge this one and fix things in the target PR.

Joozty:
I didn't some last-minute changes which I didn't test properly. Anyway, I came to the conclusion that even the modal approach is not that easy as I saw it first. The problem is that when we are doing it on the application level there may be like tiny timeframe when dashboard may be used while policies are loaded. I don't see it as a big problem. If we don't want to have it there we might need to do a bigger refactor. 


Let me explain:

1. You access dashboard.core.ac.uk
1. The terms modal is a HOC component and is rendered straight after the user is authenticated
2. We need to render /data-providers page in order to be redirected to the first available data provider
 and therefore we cannot have in Modal component the condition like `if(!dataProvider.areTermsAccepted) return null else return <>{children}<Modal /></>`. Because if we have it there the user never will be redirected to the data provider page and we will never known for which data provider should we load policies. :/ 

That was the reason why you saw the blank screen. I added the condition there and it was a mistake.

PR #475, opened 2020-09-21
Fix data provider id comparison

Joozty:


viktor-yakubiv (commented):
I am surprised you decided to add `Number` rather than remove one equality sign (`=`)

Joozty:
Eslint was complaining about that and I didn't want to suppress that rule.

PR #476, opened 2020-09-23
Fix direct access to subpages for superadmin

Joozty:


components/application/index.jsx
```diff
@@ -8,48 +9,50 @@ import ActivitySelect from './activity-select'
...
-                  />
-                ))}
-            </ActivitySelect>
-          )}
-        </SideBar>
-      )}
-      {children && <Main>{children}</Main>}
-    </Container>
-  </>
+const Application = observer(
```
viktor-yakubiv, line 12: What do you actually observe? Would not `useObserver` be better?

pages/data-providers/[data-provider-id]/content.jsx
```diff
@@ -5,9 +5,9 @@ import ContentTemplate from 'templates/content'
 
 const Content = ({ store: { dataProvider }, ...props }) => (
   <ContentTemplate
-    exportUrl={dataProvider?.works.contentExportUrl}
+    exportUrl={dataProvider?.works?.contentExportUrl}
     works={dataProvider?.works}
-    changeVisibility={dataProvider?.works.changeVisibility}
+    changeVisibility={dataProvider?.works?.changeVisibility}
```
Comment (viktor-yakubiv) [inline pages/data-providers/[data-provider-id]/content.jsx:10]: Sorry, my fault 😔

store/data-provider/index.js
```diff
@@ -32,6 +32,8 @@ class DataProvider extends Resource {
     recommender: null,
   }
 
+  @observable location = {}
```
viktor-yakubiv, line 35: Why is this needed? Does not `extendObservable` in the Resource.extend() handle this?

templates/issues/index.jsx
```diff
@@ -24,11 +24,13 @@ const IssuesTemplate = ({
         lastHarvestingDate={lastHarvestingDate}
       />
       <ErrorsOverviewCard totalCount={totalCount} errorsCount={errorsCount} />
-      <TableCard
-        totalCount={totalCount}
-        aggregation={aggregation}
-        pages={pages}
-      />
+      {pages && (
```
viktor-yakubiv, line 27: Why do you add such conditions everywhere? What is the problem?

viktor-yakubiv (approved):
Sorry, I did not carefully test the feature. Thank you for handling this 👍

templates/issues/index.jsx
Joozty: if you access the page directly the variable `pages` is undefined. This happens when you try to access a repository which is not in your organisation as a superadmin because data provider is not loaded yet.

store/data-provider/index.js
Joozty: Sadly not. I guess it has something to do with the first render. (variable is not observable at that moment)

It's needed for hiding regional warning on /deposit-compliance, i.e. if you access repository as a super admin the warning is shown becuase we don't have `location.countryCode` yet. This observable makes it rerender (hide) once it's loaded.

components/application/index.jsx
Joozty: useObserver is for making thing observable, isn't it? I just wanted to react on the already observable item.

viktor-yakubiv: No, it's a hook that accepts the function, which extracts everything you need from an observable and returns back to you.

It needs a function because of MobX can only track subscriptions from function. However, the interface is exactly the same as `useCallback`, i.e. you pass a function and get the data.

Check the example in the [documentation](https://mobx-react.js.org/observer-hook).

I asked because including a hook would produce fewer changes in git and might keep the code more readable.

viktor-yakubiv: Also, if it would be useful, check the documentation, [how MobX reacts on observables](https://mobx.js.org/best/react.html#common-pitfall-consolelog).

PR #477, opened 2020-09-23
Fix resources are loaded twice for superadmin

Joozty:
I noticed it now. Sorry I didn't include it in the previous PR. Anyway, I don't feel like this is a solution to the problem. I am not sure if it is possible to access the retrieve method from Resource instead of the one from DataProvider.

You can observe the problem when you access the repository as a superadmin. Try to open network tab and you will see resources are loaded twice.

http://core.dashboard:3000/data-providers/169/deposit-compliance

### DIFF
--- a/store/resource.js
+++ b/store/resource.js
@@ -10,6 +10,10 @@ class Resource {
 
   static defaultValues = {}
 
+  get retrieved() {
+    return !!this.id
+  }
+
   constructor(init, options) {
     this.options = { ...this.constructor.defaultOptions, ...options }
 
@@ -41,11 +45,11 @@ class Resource {
 
   retrieve(...scopes) {
     const { request } = this.options
-    if (!this.id) {
+    if (!this.retrieved) {
       return request(this.url).then(
         ({ data }) => {
           this.extend(data)
-          return this.retrieve(...scopes)
+          return Resource.prototype.retrieve.apply(this, scopes)
         },
         (error) => {
           throw error


store/resource.js
```diff
@@ -45,7 +45,8 @@ class Resource {
       return request(this.url).then(
         ({ data }) => {
           this.extend(data)
-          return this.retrieve(...scopes)
+          if (scopes.length) return this.retrieve(...scopes)
```
viktor-yakubiv, line 48: @Joozty the first option is to call `Resource.prototype.retrieve(...scopes)` explicitly. The second option is tweaking it a bit in the `DataProvider.prototype.retrieve`. Let me have a look a this.

However, on the one hand, this recursion is not obvious what is bad. On the other hand, it's quite nice, to my mind, because I implemented it. @Joozty what do you suggest here as a user of this class?

Joozty: I already expressed my opinion about it while again. In the nutshell, it took me a while to understand how it works but once I got it I liked it.

store/resource.js
```diff
@@ -10,6 +10,10 @@ class Resource {
 
   static defaultValues = {}
 
+  get retrieved() {
```
viktor-yakubiv, line 13: I decided to move this check to the separate method because it's shared with a descendent class now but I feel this method misleading: it's not clear the resource itself is loaded or the resource including all subresources is loaded (retrieved). I am keed to leave this as it is because I have no better option but I would appreciate any idea.

viktor-yakubiv (commented):
Yet another implementation. It's much longer but keeps the parent class logic unchanged. What do you think?

Joozty:
Mmm, I don't think this implementation is any better. What if I would want to load a scope? (my implementation wouldn't work too obviously). 

I thought you were going to try this `prototype.retrieve` thing. Didn't it work? 🙂

viktor-yakubiv:
I am pretty sure prototype call will work, I didn't try.

I don't mind any implementation but I want the interface to be clear as much as possible and technically speaking I don't know which option is worse.

PR #500, opened 2020-10-07
New issues design

viktor-yakubiv:
This PR is based on #321 and the [new issue design](https://www.figma.com/file/5tc4Nb6O0kqe7bamGT9vSu/Issues?node-id=113%3A12).


viktor-yakubiv:
@samuelpearce we still need to improve descriptions. If you open the preview link and navigate to any repository, you will see the full list of issues and how they are displayed now.

@Joozty I might need your help with showing the actual list of issues. @mcancellieri we might need to change the API to fit the current design.

Joozty:
@viktor-yakubiv ok let me know here or through the slack when you need any of my help.🙂

templates/issues/cards/issue.jsx
```diff
@@ -0,0 +1,134 @@
...
+  const loadMore = useCallback(async () => {
+    if (!pages) return
+
+    setLoading(true)
+    const data = await pages.slice(0, issues.length + 10)
+    await Promise.allSettled(
+      data
+        .filter(({ output }) => output == null)
+        .map((issue) =>
+          pages.request(issue.outputUrl).then((response) => {
```
viktor-yakubiv, line 22: One will kill me for this

templates/issues/cards/issue-card.module.css
```diff
@@ -0,0 +1,120 @@
...
+.card:hover {
+  box-shadow: 0 8px 16px rgba(0, 0, 0, 0.15);
+}
+
+/* Modifiers */
+
+/* It's weird this rule has a problem here */
+/* stylelint-disable string-no-newline */
+.two {
+  filter: url('data:image/svg+xml,\
```
Joozty, line 15: this is blocked by CSP in firefox. Also I had a problem with it in chrome but strangly when I removed and added these filters back it started working. 🤔 Sadly it didn't help with firefox.

image

Joozty (changes requested):

templates/issues/cards/issue-card.module.css
viktor-yakubiv: This prevents only the filter, doesn't it? The filter is not something critical for the design, it only indicated a card stack but the information is passed already, so I would not mind having this issue if it's a Firefox bug.

I am against moving these images somewhere because this is a part of styling and not something external, it's kind of shadow, a mask. I should stay in the CSS file.

I will investigate it just to investigate it but if there is no fix, I will drop it as it is.

Joozty: The problem is that that filter rule makes whole file invalid. **Maybe it's not because of CSP it may be something else**. I don't really know. Does this happen to you too?


As I said Chrome started working once I removed and added back these `.multiple` and `.two` rules but Firefox keeps showing like this. What makes it even strange is that when I open it on [vercel](https://dashboard-git-issues-redesign.oacore.now.sh/data-providers/1/issues) everything works just fine.

viktor-yakubiv: Firefox works badly on Vercel too. I will check it.

viktor-yakubiv: I did check it. If I allow `data:` in default-src, it works in Firefox. I assume that Firefox checks the protocol first and then violates the CSP when Chrome checks the MIME-type first.

I cannot say either is wrong. Also, I am afraid to add `data:` to default-src and no other -src worked. I will have a look but I am afraid I should drop the feature that is so cool.

Joozty: No, please don't drop it. I know you are against moving it into `/public` but I don't see anything wrong in doing so. It's better than just dropping whole feature because of some shitty CSP bug.

viktor-yakubiv: I don't think, it's a bug but I will find a way 🙂

Joozty:
Can you please rebase to the current master?  🙂

Joozty (approved):
Nice work :)

PR #532, opened 2020-11-24
Move ResizeObserver to iframes

Joozty:
I slightly refactored to implementation. The workflow is following:

1. Load iframe with `elementId` parameter
2. If that parameter exists and the page is loaded from within the iframe and ResizeObserver API exist
     - Listen on any resize change and update height on `elementId` element from parent
3. Else the arbitrary height is applied and user sees scrollbar (hope it never happens)

### DIFF
--- a/public/secure/login.html
+++ b/public/secure/login.html
@@ -41,4 +41,6 @@
   <a class="forgotten-password" href="/reset" target="_top">Forgotten password?</a>
 </form>
 
+
+<script src="/secure/resize.js" ></script>
 <script src="/secure/login.js" ></script>
--- a/public/secure/resize.js
+++ b/public/secure/resize.js
@@ -1,0 +1,29 @@
+function handleResizeEvent(entries, elementId) {
+  // https://stackoverflow.com/a/58701523/5594539
+  window.requestAnimationFrame(() => {
+    if (!Array.isArray(entries) || !entries.length) return
+
+    const iframe = window.parent.document.getElementById(elementId)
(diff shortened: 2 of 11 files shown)


public/secure/login.html
```diff
@@ -2,7 +2,6 @@
 <html lang="en">
 
 <meta charset="UTF-8">
-<meta name="viewport" content="width=device-width, initial-scale=1">
```
viktor-yakubiv, line 5: I see this kept in other places. Leftover?

public/secure/resize.js
```diff
@@ -0,0 +1,29 @@
+function handleResizeEvent(entries, elementId) {
+  // https://stackoverflow.com/a/58701523/5594539
+  window.requestAnimationFrame(() => {
+    if (!Array.isArray(entries) || !entries.length) return
+
+    const iframe = window.parent.document.getElementById(elementId)
+    iframe.style.height = `${window.document.body.offsetHeight}px`
```
viktor-yakubiv, line 7: Really? You can do so? It's even better 👏

By the way, did you check if some iframe attribute could deny this?

viktor-yakubiv (approved):
Great job 👍

public/secure/login.html
Joozty: I was suspecting this was causing the loop exceed but I was wrong and forgot to add it back. Thanks

public/secure/resize.js
Joozty: I think you can do so only if you are on the same origin.

PR #533, opened 2020-11-25
Escape search string

Joozty:
search function accept regex argument or converts it to regex if string is provider. If you tried search for \ it threw exception. This change fixes it.

https://sentry.io/organizations/oacore/issues/2049656723/?project=2091955&query=is%3Aunresolved

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "cssnano": "^4.1.10",
     "d3-array": "^2.8.0",
     "d3-collection": "^1.0.7",
+    "escape-string-regexp": "^4.0.0",
     "mobx": "^5.15.7",
     "mobx-react-lite": "^2.2.2",
     "mustache": "^4.0.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "cssnano": "^4.1.10",
     "d3-array": "^2.8.0",
     "d3-collection": "^1.0.7",
+    "escape-string-regexp": "^4.0.0",
```
viktor-yakubiv, line 48: Wow! This is the whole library!

```js
string
  .replace(/[|\\{}()[\]^$+*?.]/g, '\\$&')
  .replace(/-/g, '\\x2d')
```

I think writing it yourself could save you one `if` ~and one `import`~ but it's shared in 2 places.

viktor-yakubiv (approved):

Joozty:
@viktor-yakubiv Maybe it would be better if we have this in design because @oacore/search is affect to. I can imagine to have a prop like `escape=Bool` for select?

viktor-yakubiv:
The prop you suggest is very ambiguous: you can escape any chars, `"` for CSV, `\` for regex, `&` for HTML`, `%` and non-ASCII for URLs and so on. If I hear this first time, I would rather think about the last because I tend to think that input should be URL-encoded on submit.

Also, this is a big side-effect. I would duplicate this for the time being.

